PR #18, opened 2019-11-19
Run `yarn test` as well when testing creation of new project

vgrichina:


vgrichina:
@icerove did you look into test failures on CI?

icerove:
No, I don't have time today.

icerove:
I will check it on Monday

icerove:
we do not have any test script in blank_template, so `yarn test` is not effective

vgrichina:
@icerove 

> we do not have any test script in blank_template, so yarn test is not effective

1) `blank_template` would benefit from having `yarn test`
2) Tests didn't pass in React template for unclear reason.

icerove:
fix the test

PR #40, opened 2020-01-01
replace deprecated call to loadContract

mikedotexe:
Very simple deprecation replacement. I'm new to NEAR Protocol's codebase, but this seems like a quick fix.


vgrichina:
@mikedotexe thanks for contributing!

PR #54, opened 2020-01-14
Update near-shell which uses near-bindgen-as

willemneal:


blank_react_project/package.json
```diff
@@ -3,14 +3,11 @@
...
   "private": true,
   "dependencies": {
-    "assemblyscript-json": "github:nearprotocol/assemblyscript-json",
-    "bignum": "github:MaxGraey/bignum.wasm",
     "gh-pages": "^2.1.1",
     "gulp": "^4.0.2",
     "js-cookie": "^2.2.1",
-    "near-runtime-ts": "^0.5.1",
-    "near-shell": "^0.18.0",
-    "nearlib": "^0.19.1",
```
vgrichina: why remove nearlib?

blank_react_project/package.json
```diff
@@ -3,14 +3,11 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "assemblyscript-json": "github:nearprotocol/assemblyscript-json",
-    "bignum": "github:MaxGraey/bignum.wasm",
```
vgrichina: I assume these 2 are now coming from near-runtime-ts?

blank_project/package.json
```diff
@@ -15,7 +15,7 @@
         "gulp": "^4.0.2",
         "jest": "^24.9.0",
         "jest-environment-node": "^24.9.0",
-        "near-runtime-ts": "^0.5.1",
+        "near-runtime-ts": "^0.5.2",
         "near-shell": "^0.18.0"
```
vgrichina: doesn't near-shell also need update here?

blank_react_project/package.json
willemneal: It's a dependency of nearshell. Should we change that?

blank_react_project/package.json
willemneal: Yep.

blank_react_project/package.json
vgrichina: It's a dependency of near-shell, but also this project presumably uses it directly. I think it's generally good idea to list all direct dependencies in `package.json` (vs e.g. something like `near-hello` which isn't used directly here).

willemneal: Ah then should we list it as a peer dependency that way it will give a warning if the version of nearlib is incompatible with near-shell's?

vgrichina:
@willemneal do you plan to merge this?

blank_react_project/package.json
vgrichina: @willemneal I thought near-shell would use it's own version of nearlib unless we mess up semver?

icerove:
updated and no need

PR #67, opened 2020-01-22
Renames near.gitignore to .gitignore during project creation

mikedotexe:
**Note**: I'd like some guidance with folks familiar with publishing this to npm. I'm not aware of how to "test" this yet without publishing.

From my understanding (see links below) the `.gitignore` files we have created are not being created when using the generator commands like `npx create-near-app…`.
This is not good as end users may accidentally commit sensitive info, like in the `neardev` directory.

For context, here's two links discussing the missing `.gitignore` files in folks' generators.
https://github.com/npm/npm/issues/1862
https://github.com/npm/npm/issues/3763

Reading through those, there's discussion of adding a `.npmignore` file, but I don't think those threads are on track with our issue here.

My solution here is to add a new file, `near.gitignore` that will _not_ be omitted during the process of publishing to npm. Then in the `package.json` files under the `scripts` key adding `postinstall` and simply renaming it after the `npm install`. (rename `near.gitignore` to `.gitignore`)

Then when it comes to our own development, we'd like a normal `.gitignore` file, so I've created a symlink (`.gitignore` » `near.gitignore`) so we don't have to maintain two of those files.

---

While I couldn't test the `npx` command, I did locally remove the `node_modules` directory and `npm i` and see that this works, and that git uses the symlink like I hoped it would.


mikedotexe:
This person proposes the idea of "whitelisting" files. Something to think about:
https://medium.com/@jdxcode/for-the-love-of-god-dont-use-npmignore-f93c08909d8d

vgrichina:
BTW, for now I've just changed `.gitignore` files to be more specific, because they also cause test keys to not be included (breaking running tests)
https://github.com/nearprotocol/create-near-app/pull/69

mikedotexe:
> BTW, for now I've just changed `.gitignore` files to be more specific, because they also cause test keys to not be included (breaking running tests)
> #69

Interesting. So perhaps the `.gitignore` file we want the user to have after the `npx` generator command **is** going to be different than our own internal `.gitignore`
For instance we won't want this removed from the end user's file, right?
https://github.com/nearprotocol/create-near-app/pull/69/files#diff-67d2838502483fecce1727e6ebd976e3L2

icerove:
do we still need this?

mikedotexe:
> do we still need this?

I personally think it's pretty important to ship a default `.gitignore` file to the end user.
It will be hard for them to know that they need to add `neardev` themselves, for instance. And if they push up code to a public repo, that wouldn't be great.
This morning we also talked about moving all the keys to `~/.near/[some-directory]`, which would help.
I'd be happy to chat about this with folks.

mikedotexe:
So this is the most important doc I found about this:
https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package

If we want someone to use the generator command(s) `npx create-near-app path` and end up with a healthy `.gitignore` file, I think the only way is this `postinstall` option in `package.json`

OR

We should whitelist the files as described there.
One cool thing is you can try playing around with `.gitignore` and `.npmignore` and then run:

`npm pack`

This will create a tarball that can be extracted and you'll see exactly what you'd get after running the generator command.

blank_react_project/package.json
```diff
@@ -26,7 +26,8 @@
     "deploy:pages": "gh-pages -d build",
     "deploy": "npm run deploy:contract && npm run deploy:pages",
     "build": "react-scripts build && gulp",
-    "dev:deploy:contract": "near dev-deploy ./out/main.wasm"
+    "dev:deploy:contract": "near dev-deploy ./out/main.wasm",
+    "postinstall": "mv near.gitignore .gitignore"
```
vgrichina: Not sure how this should work. Like this is a script for template project, my understanding is that it won't be executed during project generation. It would be executed when someone installs generated project though. But we want `.gitignore` at time of project generation.

vgrichina:
@mikedotexe I think the easiest solution is just name file differently in repo (e.g. `gitignore` without dot) and then rename it when copying files to create new project (somewhere here https://github.com/nearprotocol/create-near-app/blob/master/index.js#L28)

Symlinks won't work on Windows I think and aren't very critical as generally there shouldn't be keys created inside of templates folders when developing this project.

mikedotexe:
> @mikedotexe I think the easiest solution is just name file differently in repo (e.g. `gitignore` without dot) and then rename it when copying files to create new project (somewhere here https://github.com/nearprotocol/create-near-app/blob/master/index.js#L28)
> 
> Symlinks won't work on Windows I think and aren't very critical as generally there shouldn't be keys created inside of templates folders when developing this project.

I like this approach. Thanks for pointing me to that function, @vgrichina. I just pushed up that change. This PR is ready for a re-review

mikedotexe:
@amgando noticed this issue here as well. https://github.com/nearprotocol/create-near-app/issues/99
Bump on this topic.
Travis is acting up and that's why it looks like it's failed:
`FetchError: request to http://shared-test.nearprotocol.com:3030/ failed, reason: socket hang up`

potatodepaulo:
This is good to go if tests are not passing simply due to flakiness as mentioned. I would re-run trying to make sure that's the case, then merge away.

PR #76, opened 2020-01-27
Update nearlib in blank_project to 0.21

janedegtiareva:


blank_project/src/index.html
```diff
@@ -76,7 +76,7 @@
           <button id="sign-out-button" >Sign-out</button>
       </div> 
   </div>
-  <script src="https://cdn.jsdelivr.net/gh/nearprotocol/nearlib@0.19.1/dist/nearlib.js"></script>
+  <script src="https://cdn.jsdelivr.net/gh/nearprotocol/nearlib@0.21.0/dist/nearlib.js"></script>
```
janedegtiareva: does dependabot do these types of deps? I guess not but it'd be nice to have something like that

janedegtiareva:
Tested that the app works after yarn start

vgrichina:
I'm getting 
```
error[E0463]: can't find crate for `std`
```

when trying to do `yarn start` on created project, see more:

https://gist.github.com/vgrichina/e3778f736a31cbdeaf5f1e3a744d9ebb

blank_project/src/index.html
vgrichina: I'll update all templates to use Parcel bundler (similar to wallet and near-chess GitPod demo). We won't have this kind of dependencies anymore when we deprecate studio.

vgrichina:
> I'm getting
> 
> ```
> error[E0463]: can't find crate for `std`
> ```
> 
> when trying to do `yarn start` on created project, see more:
> 
> https://gist.github.com/vgrichina/e3778f736a31cbdeaf5f1e3a744d9ebb

OOPS, commented on wrong PR.

PR #80, opened 2020-01-30
update project templates to use ci env

janedegtiareva:


blank_rust_project/contract/src/lib.rs
```diff
@@ -60,7 +60,7 @@ mod tests {
     fn set_get_message() {
         let context = get_context(vec![], false);
         testing_env!(context);
-        let mut contract = StatusMessage::default();
```
janedegtiareva: this is technically an unrelated change. I can pull out into diff PR if we feel like that's needed for housekeeping.

blank_project/src/config.js
```diff
@@ -30,8 +30,8 @@
             };
         case 'test':
             return {
```
vgrichina: can we make `case 'test'` fallthrough to `case 'ci'` to void maintaining duplicate config?

PR #81, opened 2020-01-30
ci for rust contract

janedegtiareva:


.travis.yml
```diff
@@ -1,9 +1,21 @@
...
   - NODE_ENV=ci-staging
 cache: yarn
+
+before_install:
+  - rustup target add wasm32-unknown-unknown
+  - nvm install 10.16
+  - nvm use 10.16
+
+install: 
+  - npm install
```
vgrichina: Should use `yarn` given we use `yarn` everywhere else (including `cache: yarn`). yarn can be installed in `before_install` using `npm install -g yarn`

janedegtiareva: done

nearmax:
@vgrichina It is preferable if you had let @janedegtiareva to merge it. She was waiting for the feedback from me. 

@janedegtiareva We need to add to near-bindgen CI the test that checks whether rust examples are not broken.

vgrichina:
> @vgrichina It is preferable if you had let @janedegtiareva to merge it. She was waiting for the feedback from me.

It's not like we cannot incorporate your feedback in follow ups. 
If I get review request and it passes CI, I assume it's good to merge if there isn't any specific comment like "DO NOT MERGE before @nearmax approval" or smth like that. It's generally much faster to move that way, especially given that I'm iterating on related stuff (GitPod).

PR #85, opened 2020-02-03
bump version in package.json to 0.5.0

janedegtiareva:


janedegtiareva:
I pushed 0.5.0 for sat internal hackathon

vgrichina:
Can we please follow the release process next time?

If `np` doesn't work, here's manual stuff that needs to happen:
https://www.notion.so/nearprotocol/Applayer-release-process-27ae5d35e1f44eb5b83184f100bf84f1#2eeff9c3fd894a96bce010dd6b419e2e
In particular there needs to be tag and release create on GitHub. This info is used by tools like dependabot, etc + generally makes it easier to find given version on GitHub.

PR #87, opened 2020-02-04
Update Rust template to work with dev-deploy

vgrichina:
This means that `yarn start` works immediately without any dance with logging in, etc


blank_rust_project/package.json
```diff
@@ -1,32 +1,32 @@
...
+    "jest-environment-node": "^24.9.0",
+    "near-bindgen-as": "^1.2.0",
+    "near-runtime-ts": "^0.5.2",
+    "near-shell": "^0.20.1",
+    "nodemon": "^2.0.2",
+    "parcel-bundler": "^1.12.4"
+  },
+  "dependencies": {
+    "nearlib": "^0.21.0",
+    "regenerator-runtime": "^0.13.3"
```
janedegtiareva: what is regenerator-runtime used for?

janedegtiareva:
Approved assuming that the build failure is a flake (I just restarted the job that failed)

blank_rust_project/package.json
vgrichina: it's used to support `async` / `await` in browsers which don't support it natively.

vgrichina: or to be more precise for browsers without support for generators, which are somewhat related concept which can be used to fake `async` / `await`

PR #94, opened 2020-02-06
Consolidate examples - use same contract api

janedegtiareva:


janedegtiareva:
https://github.com/nearprotocol/create-near-app/issues/86

PR #95, opened 2020-02-07
Refactor project generation so that contract is copied from single so…

janedegtiareva:
…urce


index.js
```diff
@@ -59,6 +59,17 @@ const doCreateProject = async function(options) {
         to: projectName
     });
 
+    // copy contract
+    const contractTargetDir = `${projectDir}/${options.rust? 'contract' : 'assembly'}`;
+    const contractSourceDir = `${__dirname}/contract/${options.rust? 'rust' : 'asc'}`;
+    console.log(`Copying contract files to new project directory (${contractTargetDir}) from source (${contractSourceDir}).`);
+    const copyContractDirFn = () => {
+        return new Promise(resolve => {
+            ncp(contractSourceDir, contractTargetDir, response => resolve(response));
```
vgrichina: This doesn’t propagate errors to caller. Use util.promisify?

icerove:
could you let the ci pass first?

janedegtiareva:
> Lgtm overall. Not sure about:
> 
> * shouldn’t there be change to index.ja in web template copying? E. g. To avoid it erroring on rust + react
> * looks like blank rust project folder still remains because of gitkeep file.

the change to generate rust + react is in flight.
We can't quite get rid of rust project yet because of readme file and package.json diffs.

index.js
janedegtiareva: Filed a reminder issue to track, will address later when rust + react are done

PR #100, opened 2020-02-13
WIP doc gen - DO NOT SUBMIT

janedegtiareva:


index.js
```diff
@@ -70,6 +70,21 @@ const doCreateProject = async function(options) {
     };
     await copyContractDirFn();
 
+    console.log('Generating documentation');
+    const { exec } = require("child_process");
```
mikedotexe: cool, I will say that in the [Windows PR](https://github.com/nearprotocol/create-near-app/pull/138) I'm adding [shelljs](https://www.npmjs.com/package/shelljs) which can use `shell.exec(…`
That may not matter, but something to keep in mind.

PR #107, opened 2020-02-20
refactor: update README

chadoh:
* strip down list of prerequisite installs to the bare minimum – the
  rest are installed via `package.json` dependencies
* explain frontend/backend configuration options with a concise table
* clean up various language

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Quickly build apps backed by the [NEAR](https://nearprotocol.com) blockchain
 Prerequisites
 =============
 
-Make sure you have a [current version of Node.js](https://nodejs.org/en/about/releases/) installed – we have tested most with version `12.x`
+Make sure you have a [current version of Node.js](https://nodejs.org/en/about/releases/) installed – we are targeting versions `10.16+`.
 
 
 Getting Started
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,71 +1,46 @@
...
+create-near-app
+===============
+
+Quickly build apps backed by the [NEAR](https://nearprotocol.com) blockchain
+
+
+Prerequisites
+=============
+
+Make sure you have a [current version of Node.js](https://nodejs.org/en/about/releases/) installed – we have tested most with version `12.x`
```
mikedotexe: I like this PR a lot. The only thing I'm hesitant about is the verbiage here with version 12. I'd like to have a more solid answer. "Yes, I promise it'll work with 10.x" where this doesn't give me that same feeling. Thoughts?

chadoh: I purposely left it vague because I'm not sure – thanks for asking about this line, @mikedotexe. 

@vgrichina do you know if we can guarantee that `create-near-app` works with a certain version of node?

vgrichina: @chadoh looks like we test with node 10.16:
https://github.com/nearprotocol/create-near-app/blob/896b023c4f9acc53c8515e9476447b90f3b82f98/.travis.yml#L13

PR #112, opened 2020-02-24
Switch deps to use a stricter range on versions

willemneal:
Before it was "^1.0.0" ~= ">=1.0.0 " and "<2.0.0",
Now it's "~1.0.0" ~= " >=1.0.0" and "<1.1.0".  This will ensure that new installs don't update any breaking changes.

I will release a`near-devtools` package to ensure that all interdependent deps are upgraded together, but before then I will be updating packages with breaking changes and don't want users to use the any of the new versions before its ready.

### DIFF
--- a/contract/asc/types.d.ts
+++ b/contract/asc/types.d.ts
@@ -1,0 +1,1 @@
+/// <reference types="near-runtime-ts/assembly/as_types" />
--- a/blank_project/package.json
+++ b/blank_project/package.json
@@ -19,14 +19,14 @@
     "gulp": "^4.0.2",
     "jest": "^25.1.0",
     "jest-environment-node": "^25.1.0",
-    "near-bindgen-as": "^1.2.0",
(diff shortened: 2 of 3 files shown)


contract/asc/types.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="near-runtime-ts/assembly/as_types" />
```
janedegtiareva: Why do we need this file in source?

willemneal: It adds the global types definitions of the runtime, e.g., `obj.JSON`, `decode<T>()`.  This could go in the tsconfig.json, but it currently points to asc's which limits the types to just the stdlib.

mikedotexe:
@willemneal I think this got forgotten. still apply?

vgrichina:
@willemneal is this still relevant or deprecated because of `near-sdk-as`?

PR #113, opened 2020-02-25
Adding a send txn in example rust app.

janedegtiareva:


blank_rust_project/src/main.js
```diff
@@ -62,14 +62,29 @@ function signedInFlow() {
...
         window.location.replace(window.location.origin + window.location.pathname);
     });
+
+    // Adding an event to change greeting button.
+    document.getElementById('change-greeting').addEventListener('click', () => {
+        setGreeting();
+    });
+}
+
+async function setGreeting() {
```
bowenwang1996: Is there a reason why message is not passed as an argument?

janedegtiareva: A good argument for that is simplifying code. It's usually safe to assume that devs know how to use basic javascript, so they don't need an example of how to pass user input from the webpage. This allows less overall code and more usefulness per line of example code (which is definitely a metric we want to optimize for)

blank_rust_project/src/main.js
```diff
@@ -62,14 +62,29 @@ function signedInFlow() {
...
+    await window.contract.set_greeting({message:'Howdy'});
+    getGreeting();
```
chadoh: Do we explain why this is necessary somewhere? I find it strange that you need to `await` two things in a row – write, then read – rather than having setter-functions return values from the blockchain. Why can't someone write a smart contract function that would allow better ergonomics here?

```js
async function setGreeting(message) {
  const { text } = await window.contract.set_greeting({ message })
  document.getElementById('speech').innerText = text
}
```

blank_rust_project/src/main.js
```diff
@@ -62,14 +62,29 @@ function signedInFlow() {
...
+}
+
+async function getGreeting() {
+    const response = await window.contract.welcome({account_id:window.accountId});
```
frol: `getGreeting` is inconsistent with the `.welcome` call. Can we align those?

janedegtiareva: done

blank_rust_project/src/main.js
janedegtiareva: This is different from setter-getter pattern (and we should probably not encourage setter-getter pattern anyways). I renamed the get fn to welcome to align better with smart contract function name.

PR #119, opened 2020-02-27
Clean up deps - remove asc compiler from rust proj package.json

janedegtiareva:


janedegtiareva:
Tested rust proj by running yarn start and loading in browser.

PR #120, opened 2020-03-02
Add new blank rust react project

janedegtiareva:


vgrichina:
@janedegtiareva what is long-term plan for this? Moving back to having separate templates for every combination?

janedegtiareva:
> @janedegtiareva what is long-term plan for this? Moving back to having separate templates for every combination?

After discussion about contract fns naming conventions:
We'd need to either get the annotations for how to expose smart contract fns externally or start doing more templating. At this point let's get the project in so that devs can use it.

vgrichina:
> After discussion about contract fns naming conventions:
> We'd need to either get the annotations for how to expose smart contract fns externally or start doing more templating. At this point let's get the project in so that devs can use it.

we also have the option to just use same names in both contract. Like if all our contract standards (e.g. https://github.com/nearprotocol/NEPs/pull/21/files) going to be based on Rust implementation of token, etc – let's just roll with Rust conventions?

janedegtiareva:
> > After discussion about contract fns naming conventions:
> > We'd need to either get the annotations for how to expose smart contract fns externally or start doing more templating. At this point let's get the project in so that devs can use it.
> 
> we also have the option to just use same names in both contract. Like if all our contract standards (e.g. https://github.com/nearprotocol/NEPs/pull/21/files) going to be based on Rust implementation of token, etc – let's just roll with Rust conventions?

I think templating is the way to go. I'll start a commonwealth discussion on whether to go with the WIP solution or something else - after a quick look at other packages it doesn't look like there's a better alternative.

janedegtiareva:
> LGTM, but I am not best person to checkout rust part, though. Please update readme to show user how to create and let's see whether it is good to keep the options like this:
> default : react + assembly
> --vanilla : plain JS + assembly
> --rust : react + rust
> --vanilla --rust : plain JS + rust

This is already there!

PR #121, opened 2020-03-04
Add send txn examples to react templates

janedegtiareva:


blank_rust_react_project/src/App.js
```diff
@@ -53,6 +59,12 @@ class App extends Component {
     console.log("after sign out", this.props.wallet.isSignedIn())
   }
 
+  async changeGreeting() {
+    console.log(this);
```
vgrichina: is this on purpose or accidental?

blank_react_project/src/App.js
```diff
@@ -53,6 +59,12 @@ class App extends Component {
     console.log("after sign out", this.props.wallet.isSignedIn())
   }
 
+  async changeGreeting() {
+    console.log(this);
```
vgrichina: is this on purpose or accidental?

blank_react_project/src/App.js
```diff
@@ -16,6 +16,7 @@ class App extends Component {
     this.requestSignIn = this.requestSignIn.bind(this);
     this.requestSignOut = this.requestSignOut.bind(this);
     this.signedOutFlow = this.signedOutFlow.bind(this);
+    this.changeGreeting = this.changeGreeting.bind(this);
```
vgrichina: why do we need these? That doesn't look right, we don't do such stuff in NEAR Wallet.

blank_rust_react_project/src/App.js
janedegtiareva: accidental. thanks!

blank_react_project/src/App.js
janedegtiareva: same here

blank_react_project/src/App.js
janedegtiareva: this is null inside changeGreeting unless this line is present.

PR #138, opened 2020-03-23
Make project cross-compatible including Windows

mikedotexe:
This PR has a few things going on. Some of them are straightforward, some will be controversial.

**Note**: this cannot be published until we [have this file](https://github.com/nearprotocol/near-shell/blob/master/commands/dev-deploy.js#L47) from `near-shell`, which I believe will be published in version 0.20.2

Most of the changes use npm packages to achieve cross-compatibility.

- The `-p` flag doesn't exist on Windows for `mkdir -p` so we're now using the [mkdirp](https://www.npmjs.com/package/mkdirp) package.
- We're using the [env-cmd](https://www.npmjs.com/package/env-cmd) package to get the `CONTRACT_NAME` to feed to parcel
- In this `nodemon` script, we can't have single quotes on Windows, so using escaped double quotes
- `build.sh` is removed in favor of `build.js` which uses the [shelljs](https://www.npmjs.com/package/shelljs) package for cross-platform compatibility.
- there is trouble with the long command from `build.sh` and found elsewhere: `RUSTFLAGS='-C link-arg=-s' cargo build --target wasm32-unknown-unknown --release --` Set those up in the `.cargo/config` file [as per documentation](https://doc.rust-lang.org/cargo/reference/config.html).

Now for the controversial part.
Currently, we have no way of knowing which version of `near-shell` the user has. All our documentation has always said to install it globally, and all our documentation and readme instructions say to run `near-shell` commands with `near <command>` and not `yarn near <command>`.

We cannot know whether the new `neardev/dev-account.env` file exists when the user runs a script that executes `dev-deploy`.

We have two options:
1. Some users experience a glaring error when they try to use this project, because they have an old version of `near-shell`.
2. We shim in an ugly stop-gap check. No developer wants to do this, but it will make the DevX for our end users completely smooth.
See https://github.com/nearprotocol/create-near-app/compare/master...hotfix/windows-compatibility#diff-92bc6953198e4be151b1a6f51edff400

Let's please remember that the goal of this PR is to allow a large pool of Windows users to join workshops, hackathons, or otherwise on their own. At the end of the day, we really need to unblock Windows users as this repository is our primary introduction to NEAR.


index.js
```diff
@@ -45,16 +45,19 @@ const doCreateProject = async function(options) {
...
 
-    console.log(`Copying files to new project directory (${projectDir}) from template source (${sourceDir}).`);
+    console.log(`Copying files to new project directory (${projectDir}) from template source (${sourceTemplateDir}).`);
     // Need to wait for the copy to finish, otherwise next tasks do not find files.
-    const copyDirFn = () => {
+    const copyDirFn = (source, dest) => {
         return new Promise(resolve => {
-            ncp(sourceDir, projectDir, response => resolve(response));
+            ncp(source, dest, response => resolve(response));
         });
```
vgrichina: While you are here let’s also fix error handling (need to reject promise if error is passed to callback).

scripts/near-shell-check.js
```diff
@@ -0,0 +1,25 @@
+/*
+ * This file is a stop-gap for upgrading projects when near-shell improves/modifies logic
```
vgrichina: Given that these are throwaway accounts, we don’t have to migrate anything IMO.

scripts/near-shell-check.js
```diff
@@ -0,0 +1,25 @@
...
+(async () => {
+    const oldDevDeployFile = 'neardev/dev-account';
+    const newDevDeployFile = 'neardev/dev-account.env';
+
+    if (fs.existsSync(newDevDeployFile)) {
+        // user already has the latest near-shell
+        return;
+    }
+
+    if (fs.existsSync(oldDevDeployFile)) {
```
vgrichina: Why mix sync and async calls?

scripts/near-shell-check.js
```diff
@@ -0,0 +1,25 @@
...
+        // user has an outdated near-shell, create necessary file
+        const readFile = (filePath) => util.promisify(fs.readFile)(filePath, 'utf8');
```
vgrichina: Note that there is API already using promises https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fs_promises_api

index.js
mikedotexe: good idea, pushed

scripts/near-shell-check.js
mikedotexe: I remember reading that these weren't in Node 10 when I did this, but now that we've bumped to 12 no reason not to.

mikedotexe: Welp, these docs are for 10 so I guess whatever I read was wrong

mikedotexe: k, pushed up a change. I do think it's important to have this migration.
A lot of end users are hobbyists and may have downloaded older versions of `near-shell` I would like to have a stop-gap for this case.

scripts/near-shell-check.js
mikedotexe: See replies on comment below

scripts/near-shell-check.js
mikedotexe: See replies below

mikedotexe:
Travis is throwing benign errors:
>Transaction nonce 4042 must be larger than nonce of the used access key 4042
and
>send_tx_commit has timed out

scripts/near-shell-check.js
vgrichina: > A lot of end users are hobbyists and may have downloaded older versions of near-shell I would like to have a stop-gap for this case.

so next `yarn dev-deploy` for them would just create new account and everything would still work fine, these are throwaway accounts in any case

index.js
```diff
@@ -45,16 +45,25 @@ const doCreateProject = async function(options) {
...
+    console.log(`Copying files to new project directory (${projectDir}) from template source (${sourceTemplateDir}).`);
     // Need to wait for the copy to finish, otherwise next tasks do not find files.
-    const copyDirFn = () => {
-        return new Promise(resolve => {
-            ncp(sourceDir, projectDir, response => resolve(response));
+    const copyDirFn = (source, dest) => {
+        return new Promise((resolve, reject) => {
+            ncp(source, dest, (err) => {
+                if (err) {
+                    console.error(`Error copying ${source} to ${dest}`, err);
```
vgrichina: feel like this results in error printed 2 times (as it will still result in exception being thrown)

blank_project/package.json
```diff
@@ -9,9 +9,9 @@
     "deploy:contract": "near deploy",
     "deploy:pages": "gh-pages -d dist/",
     "deploy": "npm run build && npm run deploy:contract && npm run deploy:pages",
-    "prestart": "npm run build:contract && npm run dev:deploy:contract",
-    "start": "CONTRACT_NAME=$(cat neardev/dev-account) parcel src/index.html",
-    "dev": "nodemon --watch assembly -e ts --exec 'npm run start'",
+    "prestart": "npm run build:contract && npm run dev:deploy:contract && node ./scripts/near-shell-check ",
```
vgrichina: this doesn't make sense – as this template will only get used to create new projects, there is nothing to migrate

blank_react_project/README.md
```diff
@@ -56,23 +56,7 @@ yarn test
...
-
-`Please enter the accountId that you logged in with: <asdfasdf>`
-
-Once you have entered your account ID, it will display the following message:
-
-`Missing public key for <asdfasdf> in default`
-`Logged in with masternode24`
-
-This message is not an error, it just means that it will create a public key for you.
+Type the command `near login` which opens a webpage at NEAR Wallet. Follow the instructions there and it will create a key for you, stored in the `/neardev` directory.
```
vgrichina: 👍 @mikedotexe @amgando really like this, let's keep working on replacing docs with tools

scripts/near-shell-check.js
```diff
@@ -0,0 +1,32 @@
...
+const fs = require('fs').promises;
+
+(async () => {
+    const oldDevDeployFile = 'neardev/dev-account';
+    const newDevDeployFile = 'neardev/dev-account.env';
+
+    // if old file exists but new file doesn't, create it
+    const checkOldDeployFile = async () => {
+        await fs.access(oldDevDeployFile)
+            .then(async () => {
```
vgrichina: why use both `await` and `.then`???

mikedotexe: removed

blank_project/package.json
mikedotexe: removed

index.js
mikedotexe: fixed

mikedotexe:
> I played with the react project a bit:
> 
> * yarn dev worked (temp account)
> * yarn test did not work (tried to use local node Test suite failed to run
>   FetchError: request to http://localhost:3030/ failed, reason: connect ECONNREFUSED 127.0.0.1:3030)
> * near login
>   yarn start - started the server, but used the temporary dev account that got created in the previous step.

we've chatted about this, but to make public that these are fixed now

.travis.yml
```diff
@@ -1,22 +1,20 @@
...
-  - nvm use 12
-  - npm install -g yarn
-
-install: 
-  - yarn install
-
 script:
-  - yarn test
-  - yarn lint
+  - npm run test
```
vgrichina: why use npm? Is it faster on CI?

vgrichina:
@mikedotexe overall looks good, but CI has serious build time increase (per single platform):

Screen Shot 2020-03-27 at 3 42 08 PM

https://travis-ci.com/github/nearprotocol/create-near-app/builds/155739057


Screen Shot 2020-03-27 at 3 42 19 PM

https://travis-ci.com/github/nearprotocol/create-near-app/builds/155986739


Potential reasons for investigation:
- npm vs yarn
- starting from node image vs starting from rust image
- are there any changes in tests themselves?

.travis.yml
mikedotexe: there was a big issue with Yarn when [running tests on Windows](https://travis-ci.community/t/build-timeout-but-actually-succeeded/2763), so part of this was to attempt to avoid it. Then I looked to see if there would be issues with caching as we've noticed `yarn` seems to work well during workshops. So [I looked into it](https://docs.travis-ci.com/user/caching/#npm-cache) and saw that it's cached by default on Travis.
At this point it seems like we could go either way. Do you want me to change it back, @vgrichina ?

mikedotexe: On this subject, this is a good time to mention that this is why we have `YARN_GPG=no` in the Travis config. I tried every which way and indeed this is an issue with Windows and Travis, as I found in a few places online. Wanted to give you the heads up on that, as it's a flag that'll be seen in some examples, too.

package.json
```diff
@@ -26,6 +26,10 @@
     "yargs": "^15.0.2"
   },
   "devDependencies": {
-    "eslint": "^6.8.0"
+    "eslint": "^6.8.0",
+    "shelljs": "^0.8.3"
+  },
+  "engines": {
+    "node": ">=10.16.0"
```
mikedotexe: note: just pushed up a commit bumping this to 12. Since this PR has been open we've shifted to requiring 12

mikedotexe:
> @mikedotexe overall looks good, but CI has serious build time increase (per single platform):
> 
> Screen Shot 2020-03-27 at 3 42 08 PM
> 
> https://travis-ci.com/github/nearprotocol/create-near-app/builds/155739057
> 
> Screen Shot 2020-03-27 at 3 42 19 PM
> 
> https://travis-ci.com/github/nearprotocol/create-near-app/builds/155986739
> 
> Potential reasons for investigation:
> 
> * npm vs yarn
> * starting from node image vs starting from rust image
> * are there any changes in tests themselves?

No changes to the tests, Vlad. I bet it does have to do with using the `node_js` language for Travis instead of `rust`, but Travis doesn't seem to have a great solution for this. I found a recipe online that used the approach here.

vgrichina:
> I bet it does have to do with using the node_js language for Travis instead of rust, but Travis doesn't seem to have a great solution for this. I found a recipe online that used the approach here.

Not sure what do you mean. We totally did use different solution before (Rust base image + nvm). Is the problem that it's not compatible with Windows? Should be possible to configure more granularly:
https://docs.travis-ci.com/user/multi-os/#python-example-unsupported-languages

mikedotexe:
after discussions with @vgrichina we're going to iterate on the Travis CI tests for Windows. Tests are still slower than he'd want. So what's the verdict, is this s show-stopper to have 4-minute CI tests?

mikedotexe:
Actually, for sure the reason why tests are slower is I've also added the missing test for the Rust React project. That was absent before and I caught that in the PR here:
https://github.com/nearprotocol/create-near-app/pull/138/files#diff-2efa5dca6871bfbe601b6171b3efbec8R8

contract/rust/build.js
```diff
@@ -0,0 +1,8 @@
+const shell = require('shelljs');
+
+shell.fatal = true; // same as "set -e"
+
+shell.cd('contract');
+// Note: see flags in ./cargo/config
```
vgrichina: @mikedotexe feel like specifying all flags here would be less confusing

vgrichina:
@mikedotexe please resolve conflicts + see comment on flags

contract/rust/build.js
mikedotexe: This is very normal in Rust, right? And if people want to run a debug version they have less to copy/paste into their terminal, yes? Ideally we'd also have the `target = "wasm32-unknown-unknown"` in the `.cargo/config` but that breaks tests because Rust hasn't addressed that issue yet.

mikedotexe: If people want to run the release build, then `yarn build:contract` is a good shortcut, but if folks want to run a debug version or other typical commands starting with `cargo…` then we're asking them to prepend an ENV variable to it? That's weird and ugly, imo. And windows will be `set RUSTFLAGs=…` while 'nix and OS X will be `RUSTFLAGS=…`
I don't see the problem here, it is conventional for the language and makes `cargo` commands shorter.

mikedotexe: Actually, it's a little worse than I thought, they have to set an env variable that persists in the command prompt session. This approach is per project and I think it's a better fit

PR #143, opened 2020-03-24
Update to near-sdk-as

willemneal:
Remove gulp and update imports to use the new name.


willemneal:
Damn these flaky tests!!
```
 FAIL  src/App.test.js

  ● Test suite failed to run

    send_tx_commit has timed out

      at JsonRpcProvider.sendJsonRpc (node_modules/nearlib/lib/providers/json-rpc-provider.js:57:27)

      at Account.retryTxResult (node_modules/nearlib/lib/account.js:53:22)

      at Account.signAndSendTransaction (node_modules/nearlib/lib/account.js:79:26)

      at Account.createAndDeployContract (node_modules/nearlib/lib/account.js:119:9)

      at LocalTestEnvironment.setup (node_modules/near-shell/test_environment.js:33:9)`
```

mikedotexe:
I think we want to keep the `gulpfile.js` in the projects for now, actually, because `near-shell`'s command `build` runs `gulp`
https://github.com/nearprotocol/near-shell/blob/master/bin/near-cli.js#L122
This is a stop-gap and we need to land on the future of `near build` but not right now imo

willemneal:
Okay, but the instructions for this doesn't say to use `near build` so I don't think it will confuse new people.
 Plus we've already removed gulp from most examples.

mikedotexe:
> Okay, but the instructions for this doesn't say to use `near build` so I don't think it will confuse new people.
> Plus we've already removed gulp from most examples.

okay you're right, I thought the individual readmes had `near build` but I was wrong. I'm good, just gotta figure out what's up with the node version and if we're cool with bumping this to node 12

PR #165, opened 2020-03-31
Travis, check multiple OSes, conditions installing npm

mikedotexe:
This adds multiple operating systems to the list of those checked by Travis.
Also has conditionals for Windows and non-Windows in order to install npm, which installs yarn.
And add `cache: cargo` as recommended here:
https://docs.travis-ci.com/user/languages/rust/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ before_install:
   - if [[ "$TRAVIS_OS_NAME" == "windows" ]]; then HOME="/c/Users/travis" && export NVS_HOME=$ProgramData/nvs && git clone --single-branch https://github.com/jasongin/nvs $NVS_HOME && source $NVS_HOME/nvs.sh && nvs --version && nvs add 12 && nvs use 12; fi
   - if [[ ! "$TRAVIS_OS_NAME" == "windows" ]]; then nvm install 12 && nvm use 12; fi
   - npm install -g yarn
+  - yarn
 script:
   - yarn test
   - yarn lint
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,21 +2,19 @@ language: rust
 rust:
   - stable
   - nightly
-
+os:
+  - linux
+  - osx
+  - windows
 env:
-  - NODE_ENV=ci
-  - NODE_ENV=ci-staging
-cache: yarn
-
+  - NODE_ENV=ci YARN_GPG=no
+  - NODE_ENV=ci-staging YARN_GPG=no
+cache: cargo
 before_install:
   - rustup target add wasm32-unknown-unknown
-  - nvm install 12
-  - nvm use 12
+  - if [[ "$TRAVIS_OS_NAME" == "windows" ]]; then HOME="/c/Users/travis" && export NVS_HOME=$ProgramData/nvs && git clone --single-branch https://github.com/jasongin/nvs $NVS_HOME && source $NVS_HOME/nvs.sh && nvs --version && nvs add 12 && nvs use 12; fi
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,21 +2,22 @@ language: rust
...
+  - osx
+  - windows
 env:
-  - NODE_ENV=ci
-  - NODE_ENV=ci-staging
-cache: yarn
-
+  - NODE_ENV=ci YARN_GPG=no
+  - NODE_ENV=ci-staging YARN_GPG=no
+cache: cargo
```
chadoh: Is there a reason to stop caching `yarn`?

.travis.yml
```diff
@@ -2,21 +2,22 @@ language: rust
...
 before_install:
   - rustup target add wasm32-unknown-unknown
-  - nvm install 12
-  - nvm use 12
+  - if [[ "$TRAVIS_OS_NAME" == "windows" ]]; then HOME="/c/Users/travis" && export NVS_HOME=$ProgramData/nvs && git clone --single-branch https://github.com/jasongin/nvs $NVS_HOME && source $NVS_HOME/nvs.sh && nvs --version && nvs add 12 && nvs use 12; fi
+  - if [[ ! "$TRAVIS_OS_NAME" == "windows" ]]; then nvm install 12 && nvm use 12; fi
```
chadoh: Why use `nvs` only for Windows?

.travis.yml
```diff
@@ -2,21 +2,22 @@ language: rust
...
   - npm install -g yarn
-
-install:
-  - yarn install
-
+  - yarn
+git:
+  autocrlf: false
```
janedegtiareva: Do we need this?  I am assuming this is for windows git? My windows git setup works without this.

chadoh: It looks like the commit right before the last one [failed](https://github.com/nearprotocol/create-near-app/pull/165/commits) with [a bunch of `CRLF` errors](https://travis-ci.com/github/nearprotocol/create-near-app/jobs/309464923)

.travis.yml
mikedotexe: good question! I can't confirm this 100% but I haven't seen multiple `cache` keys in a Travis yaml online yet. The only time I've seen the `cache: yarn` is when the "primary" (if that's the term?) `language` declared at the top if `node_js` like shown here.
https://docs.travis-ci.com/user/caching/#yarn-cache
When I was watching Travis like a hawk, I noticed the cargo stuff took forever. I thought I might as well cache cargo here:
https://docs.travis-ci.com/user/languages/rust/#dependency-management
I believe this choice was somewhat informed, and somewhat, "this is how I've seen it elsewhere online" as well as a smidge of guesswork. Perhaps we can tinker with trying multiple cache keys in a WIP branch? (Really wish I could test Travis stuff locally but looking online that doesn't seem reliable/well-supported)

.travis.yml
mikedotexe: I saw online that `nvs` is (one of) the hipster version of `nvm` for Windows.
Whatever I can do to get `npm …` commands running on Windows is the aim here. Since Gitpod packages `nvm` with the Docker image, might as well use that instead of spending cycles also getting `nvs` for other OSes.

.travis.yml
mikedotexe: Yes, Travis tests will fail for Windows without this.

mikedotexe:
I see there are "Changes requested" here but I am unclear as to what those are supposed to be. If folks have concrete suggestions or requests, please help me by clarifying a little. Thanks, friends!

.travis.yml
chadoh: fair enough

PR #166, opened 2020-03-31
test script: use shelljs async to possibly speed up tests

mikedotexe:
Trying out async on shelljs, which may make testing a little bit faster.

### DIFF
--- a/test/test-new-project.js
+++ b/test/test-new-project.js
@@ -3,28 +3,33 @@ const shell = require('shelljs');
 shell.config.fatal = true;
 
 const commands = [
-    'node index.js tmp-project',
-    'node index.js tmp-project --vanilla',
-    'node index.js tmp-project --rust',
-    'node index.js tmp-project --vanilla --rust'
+    'node index.js',
+    'node index.js --vanilla',
+    'node index.js --rust',
+    'node index.js --vanilla --rust'
 ];
 
 for (let i = 0; i < commands.length; i++) {
     // remove temporary blank project
-    shell.rm('-rf', 'tmp-project');
-    // test generating new project in new dir
-    shell.exec(commands[i], { silent: true });
-    shell.cd('tmp-project');
-    shell.env.FILE = 'package.json';
-    if (!shell.test('-e', shell.env.FILE)) {
-        shell.echo(`Couldn't find ${shell.env.FILE}`);
-        shell.exit(1);
-    }
+    shell.rm('-rf', `tmp-project-${i}`);
 
-    shell.exec('npm install');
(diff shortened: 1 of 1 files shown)


test/test-new-project.js
```diff
@@ -3,28 +3,33 @@ const shell = require('shelljs');
...
-    shell.rm('-rf', 'tmp-project');
-    // test generating new project in new dir
-    shell.exec(commands[i], { silent: true });
-    shell.cd('tmp-project');
-    shell.env.FILE = 'package.json';
-    if (!shell.test('-e', shell.env.FILE)) {
-        shell.echo(`Couldn't find ${shell.env.FILE}`);
-        shell.exit(1);
-    }
+    shell.rm('-rf', `tmp-project-${i}`);
```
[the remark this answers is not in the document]
vgrichina: given that `i` is reused between runs – it will mean that when you run tests locally (without removing dirs) – they can be passing even after breaking changes (because files from previous run are still are there) which is undesirable

test/test-new-project.js
```diff
@@ -3,28 +3,33 @@ const shell = require('shelljs');
...
+        }
+
+        shell.exec('npm install');
+        shell.exec('npm run test');
+        shell.cd('..');
 
-// remove temporary blank project
-shell.rm('-rf', 'tmp-project');
\ No newline at end of file
+        shell.rm('-rf', `tmp-project-${i}`);
```
vgrichina, line 32: might be convenient for debugging to only remove dir before starting test (faster to debug when test failed)

mikedotexe:
Yeah, I was hoping this PR might significantly speed up the testing time, but it doesn't. I'm considering closing it. Thoughts?

test/test-new-project.js
```diff
@@ -3,28 +3,33 @@ const shell = require('shelljs');
...
+        console.log('Program output:', stdout);
+        console.log('Program stderr:', stderr);
+        shell.cd(`tmp-project-${i}`);
+        shell.env.FILE = 'package.json';
+        if (!shell.test('-e', shell.env.FILE)) {
+            shell.echo(`Couldn't find ${shell.env.FILE}`);
+            shell.exit(1);
+        }
+
+        shell.exec('npm install');
```
vgrichina, line 28: @mikedotexe not directly related to this PR, but I believe that `npm install` is now redundant here (should now be executed automatically after project gets created).

janedegtiareva:
Do we want to merge this PR at some point? @mikedotexe

mikedotexe:
I'm going to close this, I don't think it's useful after all.

PR #167, opened 2020-03-31
Add as-pect example test

willemneal:


willemneal:
@mikedotexe I can't restart the test, is it possible to merge without tests passing?

blank_react_project/package.json
```diff
@@ -45,7 +45,13 @@
...
     "setupFiles": [
       "<rootDir>/src/jest.init.js"
-    ]
+    ],
+    "testEnvironment": "near-shell/test_environment",
+    "testPathIgnorePatterns": [
+      "<rootDir>/assembly/",
+      "<rootDir>/node_modules/"
+    ],
+    "runInBand": true
```
mikedotexe, line 54: 👍

mikedotexe:
While the test passed, there are two things I noticed here:
https://travis-ci.com/github/nearprotocol/create-near-app/jobs/310687591

1. Validation Warning:
  Unknown option "runInBand" with value true was found
2. UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'search' of undefined
  DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code

Not sure how long these have been around, but we should probably nip it in the bud now. Thoughts?

blank_react_project/package.json
mikedotexe: looks like they haven't added the ability for this to be in config yet, weird.
https://github.com/facebook/jest/issues/3215

PR #185, opened 2020-04-02
Add GitPod config

vgrichina:
<a href="https://gitpod.io/#https://github.com/nearprotocol/create-near-app/pull/185"></a>

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,5 @@
+tasks:
+  - init: yarn install
+github:
+  prebuilds:
+    addBadge: true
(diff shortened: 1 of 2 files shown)


vgrichina:
@mikedotexe 
```
That readme could give instructions on running node index.js --vanilla and whatnot. I think it would be difficult for someone to know what to do next for this particular repo.
```

this sounds like something that can just be in regular README.md, doesn't look specific to GitPod at all.

> I think it would be difficult for someone to know what to do next for this particular repo.

yeah, it's unclear if you don't have experience with CLI's in Node (otherwise it's clear you need to look for `bin` in `package.json` – there isn't really any custom stuff going on)

I think it makes sense to have something like `HACKING.md` in every project which would outline setup, tips&tricks, etc related to developing package. We should link relevant Node.js/webdev docs for people who don't have experience with it (this applies to http://docs.nearprotocol.com/ as well – we really assume a lot of pre-existing web/Node.js knowledge compared vs CS 101).

`README.md` can then focus on usage and just link to `HACKING.md`.

vgrichina:
@mikedotexe merged as requested change looks very independent of this PR. Makes sense to create issue to track bigger effort on having "hacking on X" docs for all our open-source projects.

PR #337, opened 2020-06-01
Use standard serde_json dependency + fix ci 

janedegtiareva:


contract/rust/Cargo.toml
```diff
@@ -9,7 +9,7 @@ crate-type = ["cdylib", "rlib"]
 
 [dependencies]
 serde = { version = "1.0", features = ["derive"] }
-serde_json = { git = "https://github.com/nearprotocol/json", rev = "1f5779f3b0bd3d2a4b0b975abc46f3d3fe873331", features = ["no_floats"] }
+serde_json = "1.0.45"
```
chadoh, line 12: I previously [got advice](https://github.com/near-examples/rust-status-message/pull/13#discussion_r416946300) to use `serde_json = "*"`

Note that this approach [can require deleting `Cargo.lock` and rebuilding when updating dependencies](https://github.com/near-examples/pow-faucet/pull/17#issuecomment-622105893). I'm not sure if it's worth mentioning this gotcha in the project README; maybe the existing GitHub discussion about it would suffice. Or maybe somewhere in docs.near.org?

janedegtiareva: We would definitely benefit from a canonical "how do we like our dependencies". We are not consistent with this in examples right now - nft has serde_json = "1.0.45"
changed this one though

chadoh: There's some ongoing [discussion happening in near-sdk-rs](https://github.com/near/near-sdk-rs/issues/164) about which notation to use

chadoh: I've been thinking I would go with the `"*"` notation until that resolves, and change everywhere if that's the decision

contract/rust/Cargo.toml
```diff
@@ -9,7 +9,7 @@ crate-type = ["cdylib", "rlib"]
 
 [dependencies]
 serde = { version = "1.0", features = ["derive"] }
-serde_json = { git = "https://github.com/nearprotocol/json", rev = "1f5779f3b0bd3d2a4b0b975abc46f3d3fe873331", features = ["no_floats"] }
+serde_json = "*"
 near-bindgen = "0.6.0"
 borsh = "0.6.1"
```
chadoh, line 14: Should we make these other dependencies use `"*"` while we're here? Looks like we should be using `near-sdk` instead of `near-bindgen`; see other updates here https://github.com/near-examples/rust-status-message/pull/13#discussion_r416946300

janedegtiareva: We would definitely benefit from a canonical "how do we like our dependencies". We are not consistent with this in examples right now - nft has serde_json = "1.0.45"
changed this one though

chadoh: Sorry, I was wondering if we should change borsh to `= "*"` while here, and otherwise make this file look like the one in my link above

That would also include an update from `near_bindgen = "0.6.0"` to `near_sdk = "0.10.0"`

contract/rust/Cargo.toml
janedegtiareva: I updated to the exact same values that Evgeny suggested in the PR you linked.

janedegtiareva: And also added an additional bump of near-sdk-rs to 0.10

chadoh:
Tests are still failing

janedegtiareva:
> Tests are still failing

Yeah, timeouts :( After a couple of rounds of restarting they are green.

PR #356, opened 2020-06-18
chore: add dual license to all package.json

chadoh:
Partially addresses https://github.com/near/devx/issues/155

I've added a `license` line to the generated `package.json` files, as they did not have this line previously. However, I did not also add `LICENSE` or `LICENSE-APACHE` files to those repositories.

Do we want to generate projects with the same dual license that we use as our NEAR standard? Or should this change only be for the root `package.json`?

### DIFF
--- a/blank_project/package.json
+++ b/blank_project/package.json
@@ -1,6 +1,7 @@
 {
   "name": "near-blank-project",
   "version": "0.0.1",
+  "license": "UNLICENSED",
   "scripts": {
     "build": "npm run build:contract && npm run build:web",
     "build:contract": "node asconfig.js",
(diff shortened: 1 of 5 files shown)


vgrichina:
> Do we want to generate projects with the same dual license that we use as our NEAR standard? Or should this change only be for the root `package.json`?

I think it makes sense to avoid imposing license for generated projects. Like these are supposed to be owned by whoever is generating, we shouldn't decide for them.

chadoh:
> I think it makes sense to avoid imposing license for generated projects. Like these are supposed to be owned by whoever is generating, we shouldn't decide for them.

Yeah, that makes sense to me. Right now, NPM complains in the newly-generated project that the `license` field isn't set – do we like that, or would we rather default to _something_, even if it's not our atypical dual license?

vgrichina:
https://stackoverflow.com/a/33755454/341267 seems like this is what we can do

chadoh:
I changed the generated projects to use `"license": "UNLICENSED"`

PR #366, opened 2020-06-24
fix: rm obsolete --validate flag from asconfig

chadoh:
This was removed in [v10.0.0](https://github.com/AssemblyScript/assemblyscript/releases/tag/v0.10.0)

We have been using this version since at least 1dd59324cfc779becf9e5ee1127925ad8b89adfc


willemneal:
Ah good catch.  Could you also update to 0.4.2 for near-sdk-as.  When installing the newest patch is probably used, but this is the one that pins the version of asc to 0.10.0.

chadoh:
Good call, @willemneal – I've bumped those versions and will merge once the build passes

PR #368, opened 2020-06-24
Update near-shell in all frontend projs to be 0.24.4

mikedotexe:


chadoh:
Note that the most recent version of near-shell was already being installed in created projects [because of the `^`](https://docs.npmjs.com/files/package.json#dependencies), so while we don't _need_ this, it's still a good idea

mikedotexe:
> Note that the most recent version of near-shell was already being installed in created projects [because of the `^`](https://docs.npmjs.com/files/package.json#dependencies), so while we don't _need_ this, it's still a good idea

yep, definitely. seeing as the prior version would blow up,I thought it would be a good idea to be explicit here

PR #369, opened 2020-06-24
frontend: change greeting "howdy" rotating list of Hello, Aloha, Bonjour

mikedotexe:
fixes https://github.com/near/devx/issues/216
It was trivial to have the greeting rotate between three different-looking words, so I did that. As you click on "change greeting" it will cycle between "Hello", "Aloha", and "Bonjour"
I opened this against the branch where we updated `near-shell`


blank_project/src/main.js
```diff
@@ -78,7 +80,8 @@ function signedInFlow() {
 }
 
 async function setGreeting() {
-    await window.contract.setGreeting({message:'Howdy'});
+    window.nearGreetingIndex = (window.nearGreetingIndex + 1) % 3;
```
chadoh, line 83: Can we use `nearGreetings.length` instead of `3`?

Also, I think we can probably get rid of the `window.` in front of all of these – that might make it easier to read, even as we add length with the `nearGreetings.length`

blank_react_project/src/App.js
```diff
@@ -60,7 +62,8 @@ class App extends Component {
   }
 
   async changeGreeting() {
-    await this.props.contract.setGreeting({ message: 'howdy' });
+    this.greetingIndex = (this.greetingIndex + 1) % 3;
```
chadoh, line 65: Same thing here – `greetings.length`

And here, we can similarly move `greetings` and `greetingIndex` outside of the component.

```js
const greetings = ['hello, 'aloha', 'bonjour'];
var greetings = 0

class App extends Component {
...
```

chadoh: But wait! Here and above, we're not initializing `greeting` from the value in the contract! If you refresh the page, you will see "Hello" again, even if the contract state says "Aloha" – right?

mikedotexe: > But wait! Here and above, we're not initializing `greeting` from the value in the contract! If you refresh the page, you will see "Hello" again, even if the contract state says "Aloha" – right?

Not sure what you mean, there is no "Hello" in the frontend at all. What happened when you tried to refresh it?

chadoh: You're not using the value returned by `welcome()` to figure out the initial `greetingIndex`. That means that your greeting index might not match what's stored on-chain. This leads to two problems:

1. if you refresh the page when the contract is already storing "aloha", then clicking the button will not do anything
2. now that this frontend logic is more complex, it adds confusing coupling between the frontend & backend; I think it makes this app code harder to understand

blank_project/src/main.js
mikedotexe: done

mikedotexe:
pushed up the variable changes and greetings.length @chadoh

blank_project/src/main.js
```diff
@@ -5,6 +5,8 @@ import getConfig from "./config"
 
 let nearConfig = getConfig(process.env.NODE_ENV || "development");
 window.nearConfig = nearConfig;
+const nearGreetings = ['Hello', 'Aloha', 'Bonjour'];
+let nearGreetingIndex = 0;
```
chadoh, line 9: I think this comment still applies, right?

image

mikedotexe: yes, in the same way that if someone refreshed create-near-app now the "change greeting" button would do nothing.
what would you like to see here? a map keeping track of what came back from the chain and updating the index?

chadoh: > in the same way that if someone refreshed create-near-app now the "change greeting" button would do nothing

great point

PR #372, opened 2020-06-24
restyle frontends; rm duplicate templates; reorganize code; simplify contracts

chadoh:
Fixes near/devx#215

This is a large change. I recommend looking at individual commits & their messages to understand it.

### before you sign in

light mode:

Welcome to NEAR - light

dark mode:

Welcome to NEAR - dark

---

### after you sign in

light mode and dark mode work as shown above; here's dark:

changing Hello to Bonjour


mikedotexe:
Looks a ton better. I'd like either less text or perhaps putting the "Hello …" and "Howdy …" / "Aloha …" to be right next to the button. I personally didn't even see it change because my eyes were on the button. That may be one thing that the originally design had correct.
Going to definitely throw an approval on this though, as I know there are design talks taking place.

PR #386, opened 2020-07-01
feat: rapid development for create-near-app itself

chadoh:
Previously, in order to make changes to one of the templates in
create-near-app, the development process looked something like:

1. use create-near-app to generate a new app
2. fix bugs or make improvements in the new app
3. backport the changes to create-near-app proper

This is error-prone and slow, and makes it hard for people new to
create-near-app to submit improvements.

Instead, this change makes it possible to switch right into a template
directory and start working:

1. `cd templates/vanilla` or `cd templates/react`
2. run `yarn`, `yarn test`, `yarn dev`, or whatever. Make changes.

Now your changes are immediately reflected to the correct file in
the `create-near-app` repo with no backporting. And when you actually
use `create-near-app` to create a new project, the symlinks and project
build files (like node_modules) do not get copied over. To aid
troubleshooting these more complex copy functions, a new
`--very-verbose` switch has been added, which prints out every specific
file that is copied and skipped.

This also sets up a `files` directive in `package.json` and a new
`.npmignore` file in order to safely ignore these symlinks, to avoid
causing problems for Windows users who don't have symlinks & git
configured properly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ tmp-project
 .env.test.local
 .env.production.local
 .idea
+/templates/**/yarn.lock
 
 npm-debug.log*
 yarn-debug.log*
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,7 @@
+templates/*/assembly
+templates/*/contract
+templates/*/package.json
+templates/*/src/assets
+templates/*/src/config.js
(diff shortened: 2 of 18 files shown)


chadoh:
@janedegtiareva can you update the new `Contributing` section in the README with Windows instructions?

PR #387, opened 2020-07-02
feat: default to vanilla JS instead of React

chadoh:
Our user testing showed that not everyone was excited about defaulting to React.

Furthermore, defaulting to Vanilla and changing the argument syntax makes it easier to support new frontend templates (and other smart contract languages) down the road.


mikedotexe:
Let's remember this page in docs, too
https://docs.near.org/docs/quick-start/create-near-app

index.js
```diff
@@ -9,27 +9,6 @@ const spawn = require('cross-spawn');
...
 const which = require('which');
 
-const exitOnError = async function(promise) {
-    try {
-        await promise;
-    } catch (e) {
-        console.log('Error: ', e);
-        process.exit(1);
-    }
-};
```
chadoh, line 19: I removed this function, since I believe this is the default behavior. If JavaScript encounters an error, it stops executing and logs the error.

index.js
```diff
@@ -9,27 +9,6 @@ const spawn = require('cross-spawn');
...
-
-const createProject = {
-    command: '$0 <projectDir>',
-    desc: 'create a new blank react project',
-    builder: (yargs) => yargs
-        .option('projectDir', {
```
chadoh, line 25: Listing this as an `option` made it show up in the help output alongside `--rust` and `--vanilla` options, which is not correct. I've moved a streamlined version of this `.command` right into the `yargs` block at the bottom.

index.js
```diff
@@ -131,17 +110,21 @@ Happy hacking!
...
+        default: 'vanilla',
     })
-    .option('rust',{
-        desc: 'use rust for smart contract',
-        type: 'boolean',
-        default: false
+    .option('contract', {
+        desc: 'language for smart contract',
+        choices: ['assemblyscript', 'rust'],
+        default: 'assemblyscript'
```
chadoh, line 125: I've renamed this from `asc` to `assemblyscript` everywhere for the purpose of making the `--help` output easier to understand without insider knowledge.

mikedotexe:
`near-shell` tests would also need to be changed
https://github.com/near/near-shell/blob/master/test/test_contract.sh#L5
https://github.com/near/near-shell/blob/master/test/test_account_operations.sh#L4

mikedotexe:
and readme here:
https://github.com/near-examples/rust-high-level-cross-contract

mikedotexe:
worth combing through the Google Slides to see if those need to be changed, too

mikedotexe:
and here:
https://github.com/nearprotocol/workshop

chadoh:
Thanks, @mikedotexe! I created draft PRs in the repos you noted above (see linked PRs above), except for the one in `near-examples`. I noticed that we have a few simple references to `create-near-app` in our examples; I'll clean all of these up and push right to `master` once we get this one merged. See below:

Screen Shot 2020-07-02 at 17 28 02

I also looked through a bunch of our slides and updated where needed. Not sure I caught _all_ of them, but I caught a bunch in the workshops we'll likely do over & over, such as the intros to Rust and AssemblyScript.

chadoh:
@ashleynear @potatodepaulo I know y'all didn't scope this for near/devx#220, but I did it now anyhow because I think it's a necessary prerequisite to having someone complete near/bounties#17 (add a Vue template to create-near-app) or similar bounties in the future. Why? If you were a community contributor, would you feel confident suggesting changes to the shape of the flags used by `create-near-app`? Probably not. So when you see that right now we only have two flags, `--vanilla` and `--rust`, you'd probably add `--vue`. And that's going to result in some really annoying code throughout the `index.js` file.

But with this change, adding `--frontend=vue` is going to be very straightforward. A community contributor will be able to make improvements with confidence.

README.md
```diff
@@ -15,17 +15,16 @@ Make sure you have a [current version of Node.js](https://nodejs.org/en/about/re
...
-* backend: do you want your contracts to be written in **Rust** or **AssemblyScript** (a dialect of TypeScript)?
+    npx create-near-app new-awesome-project
 
-| command                                                | frontend   | backend        |
-| ------------------------------------------------------ | ---------- | -------------- |
-| `npx create-near-app new-awesome-app`                  | React      | AssemblyScript |
-| `npx create-near-app --vanilla new-awesome-app`        | vanilla JS | AssemblyScript |
-| `npx create-near-app --rust new-awesome-app`           | React      | Rust           |
-| `npx create-near-app --vanilla --rust new-awesome-app` | vanilla JS | Rust           |
+This will use [npx](https://blog.npmjs.org/post/162869356040/introducing-npx-an-npm-package-runner) to create a project with a **vanilla JavaScript** frontend and an [AssemblyScript](https://docs.near.org/docs/roles/developer/contracts/assemblyscript) smart contract.
```
janedegtiareva, line 22: Should we consider changing to plain JavaScript? I think that's a more standard language, although maybe less fun. The benefit of being understandable regardless of cultural background is definitely high.

chadoh: I really like this direction, but if we change it here I wonder if we should also change it everywhere else. The help text when you type `create-near-app --help` mentions "vanilla"; the template name is `vanilla` – where do we change it to `plain` and where do we keep `vanilla`? The current approach at least has the virtue of consistency. Let's keep as-is for now and think about this more before we change it.

Issue #374, opened 2020-06-26
Offer to install Rust automatically


vgrichina:
Instead of merely sending users to install Rust based on docs (https://docs.near.org/docs/roles/developer/contracts/near-sdk-rs#pre-requisites) we can detect whether Rust is installed and offer to do it automatically otherwise. 

Workflow is roughly:
1. Check if `rustup` is available, offer to install otherwise.
2. Check if `wasm32-unknown-unknown` target is available, offer to install otherwise.

Offer to install can be something like this:

```
To build Rust smart contracts you need to install Rust compiler for WebAssembly. We can run following command to do it:

     rustup target add wasm32-unknown-unknown

Continue with installation? [yes]:
```

mikedotexe:
really good idea

PR #373, opened 2020-06-26
Rust installation instructions

janedegtiareva:

PR #389, opened 2020-07-03
feat: improve README in generated app

### LINKED ISSUE #296: yarn dev fails rust react
yarn dev gives the following error for rust react app

```
13 info lifecycle new-awesome-app@0.1.0~dev:deploy:contract: Failed to exec dev:deploy:contract script
14 verbose stack Error: new-awesome-app@0.1.0 dev:deploy:contract: `near dev-deploy --wasmFile ./contract/res/status_message.wasm`
14 verbose stack Exit status 1

cp: no such file or directory: ./target/wasm32-unknown-unknown/release/status_message.wasm
```

chadoh:
fixes near/devx#213
fixes #296 

* move to `common/contracts/*/README.md`, as README files have contract-language-specific info but no frontend-specific info
* in index.js, move README.md from contract folder to root folder
* use https://github.com/near-examples/guest-book as starting point
* rename `vanilla/src/main.js` to `index.js` for uniformity with react template
* extend `near-blank-project` replacement in index.js to also replace all occurrences in README.md
* replace `npm` references with `yarn`, if the person running `create-near-app` has it installed


chadoh:
thanks @ashleynear, updated!

chadoh:
@mikedotexe @janedegtiareva tagging y'all for a code review (as opposed to the copy review that Ash did)

common/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+Quick Start
+===========
+
+To run this project locally:
+
+1. Prerequisites: Make sure you have [Node.js] ≥ 12 installed
+2. Install dependencies: `npm install`
+3. Run the local development server: `npm run dev` (see `package.json` for a
+   full list of `scripts` you can run with `npm`)
```
janedegtiareva, line 15: We have different prereqs for asc and rust (rust requires some extra installation steps). Does it still make sense to put this in common?

chadoh: Would we need different steps if we implemented #374?

chadoh: This is related to #373, right? 

My thought was that the READMEs seemed 95% similar, so it becomes quite a burden to maintain two. If necessary, we can, but I would rather prioritize #374 and keep the READMEs maintainable, if that makes sense to you, @janedegtiareva

mikedotexe: > The backend code lives in the `/assembly` folder
different for Rust
> The smart contract code gets tested with the [asp] command
different for Rust

chadoh: Yep, three things are different right now. These are currently being replaced when a Rust app is created; see https://github.com/near/create-near-app/pull/389/files#diff-168726dbe96b3ce427e7fedce31bb0bcR93-R95

I think this seems more maintainable than having N*M duplicate READMEs, where `N=number of frontend templates` and `M=number of contract languages`, but that's exactly what I want feedback on

(Would we need N*M? Maybe we only need the same number as contract languages)

Pros/cons for current approach:

* pro: no one will accidentally change one kind of README while leaving the others outdated
* con: the `replaceInFiles` calls could result in unexpected typos/errors in the Rust version
* con: someone might add AssemblyScript-specific stuff to the README and forget to add new `replaceInFiles` calls, resulting in errant Rust READMEs

When thinking about this, I want to aim for the implementation that results in the easiest maintainability for 90% of changes. I want something that's mostly fool-proof, that could be maintained by a community of volunteers.

Am I weighting the "pro"  too heavily?

potatodepaulo: I think that the pro might be weighed a little too heavily. I wouldn't think to update a set of functions if I were just updating READMEs. I think it's more likely to confuse than help. It makes sense to me to have just two readmes specific to the backend code that explain the project, and then you can have a separate readme for how to run the frontend in `src/` assuming they are necessary. Sure it means more READMEs, but they should be specific enough to each. This seems like one step into over-optimization.

common/contracts/rust/README.md
```diff
@@ -1,9 +1,107 @@
...
+---------------
+
+One command:
+
+    npm run deploy
+
+As you can see in `package.json`, this does two things:
+
+1. builds & deploys smart contract to NEAR TestNet
+2. builds & deploys frontend code to GitHub using [gh-pages]. This will only work if the project already has a repository set up on GitHub. Feel free to modify the `deploy` script in `package.json` to deploy elsewhere.
```
mikedotexe, line 88: awesome to add this thing about gh-pages

PR #398, opened 2020-07-08
fix: catch error on setGreeting

### LINKED ISSUE #70: add some UI that does error handling for fn calls
Right now the example app does not notice errors on rpc calls and does not report them to the user

### LINKED ISSUE #391: Can't update template twice in new generated template
Screen Shot 2020-07-06 at 3 50 22 PM

Then update to add random characters and hit save. It goes through.

If I continue updating, the `save` button doesn't unlock a second time

Screen Shot 2020-07-06 at 3 50 37 PM

### LINKED ISSUE #397: App gets stuck showing progress when error occurs
Screen Shot 2020-07-07 at 10 43 27 PM

I think it happens for any error, but in my case it was stale key for localhost.

chadoh:
I recommend reviewing the code [with indentation changes hidden](https://github.com/near/create-near-app/pull/398/files?diff=unified&w=1)

Fixes #391 by moving button disabling to vanilla/src/main.js, as this logic interferes with React's state management and breaks the React app

Fixes #397 (also fixes #70) by adding error handling to the setGreeting and moving form reenabling to the `finally` clause of the promise resolution

### DIFF
--- a/common/frontend/utils.js
+++ b/common/frontend/utils.js
@@ -35,17 +35,23 @@ export async function onSubmit(event) {
   // disable the form while the value gets updated on-chain
   fieldset.disabled = true
 
-  // make an update call to the smart contract
-  await contract.setGreeting({
-    // pass the value that the user entered in the greeting field
-    message: greeting.value
-  })
-
-  // re-enable the form
-  fieldset.disabled = false
-
-  // disable the save button, since it now matches the persisted value
-  document.querySelector('form button').disabled = true
+  try {
+    // make an update call to the smart contract
+    await contract.setGreeting({
+      // pass the value that the user entered in the greeting field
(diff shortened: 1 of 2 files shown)


common/frontend/utils.js
```diff
@@ -38,16 +38,22 @@ export async function onSubmit(event) {
   fieldset.disabled = true
 
   // make an update call to the smart contract
-  await contract.setGreeting({
```
vgrichina, line 41: @chadoh why change from `async` / `await` to returning promise explicitly?

chadoh: ...because I hadn't known about `try...catch...finally`

I've updated to use this syntax, thanks for teaching me!

PR #399, opened 2020-07-08
Add a message explaining what's happening while app is starting

janedegtiareva:
Fixes near/devx#212


templates/react/packagejsons/asc/package.json
```diff
@@ -11,7 +11,7 @@
     "deploy:pages": "gh-pages -d dist/",
     "deploy": "npm run build && npm run deploy:contract && npm run deploy:pages",
     "prestart": "npm run build:contract && npm run dev:deploy:contract",
```
ashleynear, line 13: The app is starting! It will automatically open in your browser when ready.

PR #401, opened 2020-07-08
Update dependabot config: new format + path structure

janedegtiareva:


.dependabot/config.yml
```diff
@@ -1,44 +1,32 @@
...
-          update_type: "all"
+version: 2
+updates:
+- package-ecosystem: npm
+  directory: "/templates/react/packagejsons/rust"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10
+- package-ecosystem: cargo
+  directory: "contract/rust"
```
chadoh, line 9: This is not correct anymore. The `Cargo.lock` file lives in `/common/contracts/rust`

janedegtiareva:
> Is this the change it recommends when you click this button?
> 
> Screen Shot 2020-07-08 at 19 43 09
> 
> If so, I'd like to see this PR add a new way to handle auto-merging, as it is not supported by the GitHub-native Dependabot (boo, GitHub! 👎 ), and I find it indispensable
> 
> Screen Shot 2020-07-08 at 19 44 13
Oh I didn't realize that this feature is no longer supported. Let's not upgrade to the new version then.

PR #404, opened 2020-07-09
README.md: adding a note about `yarn create`

robert-zaremba:
README.md clearly describes how to use `npx`, but doesn't provide an example with Yarn.
This PR adds a note about Yarn usage in Getting Started.


chadoh:
Thanks, @robert-zaremba !

PR #418, opened 2020-07-23
refactor: rm devnet, add explorer url

janedegtiareva:


janedegtiareva:
making a new PR due to git not cooperating to update commit msg

PR #425, opened 2020-07-28
feat: add asbuild support

willemneal:
- [x] replaces `asconfig.js` with `asconfig.json`, which is used in the newest version of assemblyscript.
- [x] adds asbuild, which has the binary, `asb`, to build release `npm run build:contract` and `npm run build:contract:debug`.

~Now binaries are in `build/<target>/` so `near deploy` requires `--wasmfile build/<target>/...wasm`~ Fixed asbuild so if you provide a binary file in the config or cli it will not output to `build/<target>/`


chadoh:
Are the test failures legit? They don't seem like transient errors

willemneal:
Yeah I'm fixing it.  Have to update asbuild.

package.json
```diff
@@ -55,6 +55,7 @@
     "eslint-plugin-react": "7.x",
     "eslint-plugin-react-hooks": "2.x",
     "husky": "^4.2.5",
+    "near-sdk-as": "^1.0.0-3",
```
chadoh, line 58: why is it now a dependency in the root package.json?

.gitignore
```diff
@@ -31,3 +31,5 @@ tmp-project
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+build/
```
chadoh, line 35: Is this folder being used somewhere? I don't see anything referencing it

willemneal: It's the default directory for asbuild.  I thought having it would be useful if in the future we move to using its defaults instead of `out/main.wasm`

chadoh: let's add it when we start using it then; I still think the root directory would be the wrong place for such a folder in this project, though

package.json
```diff
@@ -71,7 +72,9 @@
       "@commitlint/config-conventional"
     ],
     "rules": {
-      "body-max-line-length": [0]
+      "body-max-line-length": [
+        0
+      ]
```
chadoh, line 77: I doubt there should be any changes to the root `package.json` for this change

willemneal: It happens everytime I run yarn so I figured why the hell not.

chadoh:
@willemneal what’s the status on this?

willemneal:
@chadoh forgot to merge 🥇

PR #433, opened 2020-08-04
Use near-cli instead of near-shell

mikedotexe:
We're now using
https://www.npmjs.com/package/near-cli
and have a deprecation warning at the top of:
https://www.npmjs.com/package/near-shell

Also the Github has been renamed to:
https://github.com/near/near-cli


chadoh:
Thanks, @mikedotexe!

PR #447, opened 2020-08-13
(fix): add .strict() to yargs to catch mistyped options

### LINKED ISSUE #409: creating app requires yarn
In `index.js` we perform two checks, to see if the user has `npm` and if they have `yarn`

https://github.com/near/create-near-app/blob/ad6da2e7969e0d8585f3c41526bea28642bd6899/index.js#L104-L105

If they have either of them, we automatically install dependencies

This check doesn't work; it always thinks the machine has `yarn` installed. This means that when it tries to run `yarn install`, an error is thrown, which makes people think they **need** to install yarn. This is unnecessary and can be quite painful for new JavaScript developers.

luciotato:
- if the user mistypes an option, for example `create-near-app --frrontend react myapp`, the option was ignored and the app created anyway with the default options (vanilla).
By adding `.strict()`, yargs emits an error

fix: work even if yarn is not installed, fixes #409

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ tmp-project
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+/.vs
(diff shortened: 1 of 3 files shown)


chadoh:
Thanks for this!

PR #452, opened 2020-08-17
feat(templates): add vue template

luciotato:
new template in templates/vue3
fixes https://github.com/near/bounties/issues/17

also small fixes/additions to
.gitignore


chadoh:
@luciotato can you rebase this branch or otherwise push a new (potentially empty) commit, please? Travis CI was configured to run on branches but not Pull Requests, and so it never ran the tests for this branch. Sorry about that!

templates/vue3/src/App.vue
```diff
@@ -0,0 +1,160 @@
...
+const nearConfig = getConfig(process.env.NODE_ENV || 'development')
+
+console.log(`networkId:${nearConfig.networkId} CONTRACT_NAME:${nearConfig.contractName}`);
+
+import { login, logout, onSubmit } from './utils'  
+import './global.css'
+
+import Notification from './components/Notification.vue'
+
+    export default {
```
chadoh, line 100: should this be outdented?

templates/vue3/src/App.vue
```diff
@@ -0,0 +1,160 @@
...
+            }
+        },
+        name: 'App',
+        components: {
+            Notification
+        },
+        data: function () {
+            return {
+                savedGreeting: "Hello stranger",
+                newGreeting: "Hello stranger",
```
chadoh, line 113: are these defaults? seems strange to default them to “Hello stranger” when the default the contract will return is just “Hello”. I’d rather they defaulted to empty strings or null

templates/vue3/src/App.vue
```diff
@@ -0,0 +1,160 @@
...
+                console.log("calling utils.login")
+                login()
+                this.retrieveSavedGreeting()
+            },
+
+            logout: logout,
+
+            retrieveSavedGreeting()  {
+                //retrieve greeting
+                var self=this
```
chadoh, line 135: If you change the `retrievSavedGreeting` line to use an arrow function, I think you can get rid of the `var self=this`

```js
retrieveSavedGreeting: () => {
  ...

File: templates/vue3/src/App.vue [hunk 4]
```diff
@@ -0,0 +1,160 @@
...
+            // fire frontend-agnostic submit behavior, including data persistence
+            // look in utils.js to see how this updates data on-chain!
+            await onSubmit(event) //gets data directly from HTML elements
+
+            // update upper `greeting` message with persisted value
+            this.savedGreeting = this.newGreeting;
+            this.updating = false;
+
+            this.notificationVisible = true; //show notification
+            //Notification auto-hides after 10s because tag <aside>.- see: global.css
```
chadoh, line 154: without a timed removal from the dom or setting `display:none`, do multiple updates in a row open the notification again? in the other templates there’s a `setTimeout` here

templates/vue3/src/main.js
```diff
@@ -0,0 +1,15 @@
+import Vue from 'vue'
+import App from './App.vue'
+
+Vue.config.productionTip = false
```
chadoh, line 4: What’s this? Should it be configured with an environment variable?

templates/vue3/vue.config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+    chainWebpack: config => {
+        config
+            .plugin('html')
+            .tap(args => {
+                args[0].title = "My NEAR Vue App";
```
chadoh, line 6: should this also use `near-blank-project` and be added to that `replaceInFiles` block in `index.js`?

.gitignore
```diff
@@ -2,9 +2,12 @@
 
 # dependencies
 /node_modules
+node_modules
```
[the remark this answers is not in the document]
luciotato: I'll remove it

.gitignore
```diff
@@ -2,9 +2,12 @@
...
 /.pnp
 .pnp.js
 /out
+/tmp-project
+out
```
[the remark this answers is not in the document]
luciotato: I'll remove them

.gitignore
```diff
@@ -2,9 +2,12 @@
 
 # dependencies
 /node_modules
+node_modules
 /.pnp
 .pnp.js
 /out
+/tmp-project
```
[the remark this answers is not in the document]
luciotato: I'll remove it

common/frontend/config.js
```diff
@@ -1,4 +1,4 @@
-const CONTRACT_NAME = process.env.CONTRACT_NAME ||'near-blank-project'
+const CONTRACT_NAME = process.env.CONTRACT_NAME || process.env.VUE_APP_CONTRACT_NAME || 'near-blank-project'
```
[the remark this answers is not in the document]
luciotato: It's a Vue3 security issue. I tried several possible solutions and this was the most elegant. 
Vue3, as a security measure, **only imports process.env variables starting with VUE_APP_** to avoid accidentally exposing a private key. See https://cli.vuejs.org/guide/mode-and-env.html#environment-variables

index.js
```diff
@@ -149,7 +149,7 @@ const opts = yargs
   .example('$0 new-app', 'Create a project called "new-app"')
   .option('frontend', {
     desc: 'template to use',
-    choices: ['vanilla', 'react'],
+    choices: ['vanilla', 'react', 'vue3'],
```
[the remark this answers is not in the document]
luciotato: I´ll change it to vue.

templates/vue3/.gitpod.yml
```diff
@@ -0,0 +1,6 @@
+tasks:
+  - init: yarn
+    command: yarn dev
+ports:
+  - port: 8080
+    onOpen: open-browser
```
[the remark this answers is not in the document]
luciotato: hmmm... it's not the same port... vue uses 8080, the other 2 templates use parcel, port 1234

templates/vue3/packagejsons/assemblyscript/package.json
```diff
@@ -0,0 +1,85 @@
+{
+  "name": "my-vue-project",
```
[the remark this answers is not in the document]
luciotato: yes, you're right! thanks

templates/vue3/packagejsons/assemblyscript/package.json
```diff
@@ -0,0 +1,85 @@
...
+    "build:web": "node copy-dev-account.js && vue-cli-service build",
+    "dev:deploy:contract": "near dev-deploy",
+    "deploy:contract": "near deploy",
+    "deploy:pages": "gh-pages -d dist/",
+    "deploy": "npm run build && npm run deploy:contract && npm run deploy:pages",
+    "prestart": "npm run build:contract && npm run dev:deploy:contract",
+    "start": "echo The app is starting! It will automatically open in your browser when ready && npm run serve",
+    "dev": "nodemon --watch assembly -e ts --exec \"npm run start\"",
+    "test": "asp --nologo && npm run build:contract && jest test --runInBand",
+    "serve": "node copy-dev-account.js && vue-cli-service serve",
```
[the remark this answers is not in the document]
luciotato: I recommend using `vue-cli-service` to make the best experience possible to people coming from Vue into NEAR. Vue has those functions integrated, using `vue-cli-service` is what vue programmers would expect. https://cli.vuejs.org/guide/cli-service.html#using-the-binary

templates/vue3/src/App.vue
```diff
@@ -0,0 +1,160 @@
+<template>
+    <div id="root">
```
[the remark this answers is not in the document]
luciotato: No, I'll fix that. I had 4 spaces as default in vscode

templates/vue3/src/App.vue
luciotato: No. I'll fix that.

templates/vue3/src/App.vue
luciotato: I'll change it to "Hello"

templates/vue3/src/App.vue
luciotato: You're right, I'll use an arrow function

templates/vue3/src/App.vue
luciotato: There's a single <aside> for notification, multiple updates in a row do not show more than one notification, but you're right, the `setTimeout `is required in order to restart the keyframe animation . **I'll restore the setTimeout.**

templates/vue3/src/main.js
luciotato: It's standard to set it to false. Just stops an annoying "tip" from appearing every time you run. https://stackoverflow.com/questions/41743926/disable-development-mode-warning-in-vuejs

templates/vue3/vue.config.js
luciotato: You're right, I'll do that.

luciotato: Found a better way during App creation. I'll remove the vue.config.js

templates/vue3/.gitpod.yml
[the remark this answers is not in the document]
chadoh: ah, I missed that detail, let's keep it here then

templates/vue3/src/App.vue
chadoh: After someone has changed the greeting and they refresh the page, will setting this to "Hello" cause "Hello" to flash briefly on the screen before the updated greeting loads? I'd rather have an empty string or `...` show up. Seeing "Hello" briefly seems like it could cause confusion.

templates/vue3/src/App.vue
chadoh: Yes, this is sort of a fake notification, as we wanted to avoid overcomplicating it, but maybe this was a bad way to simplify it. Do you think there's an easy way to insert a new `<aside>` on every form submission? If you would rather do that, I think this might be an ok place for the Vue template to break with the patterns of the other two.

common/frontend/config.js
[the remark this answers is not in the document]
chadoh: Maybe we should move `config.js` into each individual template, rather than putting it in `common/frontend`

My medium-term goal is to obsolesce this file entirely by putting sensible defaults in `near-ap-js` and using `.env` for `CONTRACT_NAME`, at which point we'll be able to get rid of the largely-duplicate `config.js` files in each template directory quite easily.

Are you up for doing this move now?

templates/vue/asconfig.js
```diff
@@ -0,0 +1,11 @@
...
+
+compile('assembly/main.ts', // input file
+  'out/main.wasm',    // output file
+  [
+    //   "-O1",          // Optional arguments
+    '--debug',
+    '--measure',         // Shows compiler runtime
+  ], {
+    verbose: true     // Output the cli args passed to asc
+  })
```
chadoh, line 11: While you were working, `asconfig.js` was deprecated! Please update to `near-sdk-as@1.0.x`. This will let you use `asconfig.json` instead of `asconfig.js`, and you'll need to update some of the `package.json` scripts to use a new `asb` command to build the contract. Use #425 as a guide. 

Also, just to let you know, sometime soon I want to reorganize the assemblyscript contract folder so that we don't need this assemblyscript-specific stuff in the frontend project root directory, and we'll also need fewer exceptions in the `index.js` file (no more `assembly` folder vs a `contract` folder, for example). If that had been in place before you started, then the change from `asconfig.js` to `asconfig.json` would not have required any changes in your PR.

templates/vue/src/components/Notification.vue
```diff
@@ -0,0 +1,32 @@
+﻿<template>
+    <div class="notification">
```
chadoh, line 2: Please fix indentation here, too

templates/vue/src/main.js
```diff
@@ -0,0 +1,15 @@
+import Vue from 'vue'
+import App from './App.vue'
+
+Vue.config.productionTip = false
+
+import { initContract } from './utils'
+
+window.nearInitPromise = initContract()
+    .then(() => {
```
chadoh, line 9: this file should also use 2 space indents

templates/vue/src/__mocks__/fileMock.js
```diff
@@ -0,0 +1,3 @@
+// NOTE: This is used to mock resource imports in JSX for tests
```
chadoh, line 1: JSX?

templates/vue/src/App.vue
```diff
@@ -0,0 +1,165 @@
...
+
+    this.notificationVisible = true; //show new notification
+
+    // remove Notification again after css animation completes
+    // this allows it to be shown again next time the form is submitted
+    setTimeout(() => {
+      this.notificationVisible = false
+    }, 11000)
+    }
+  }
```
chadoh, line 161: Is this a minor indentation error or an extra closing bracket?

templates/vue/packagejsons/rust/package.json
```diff
@@ -0,0 +1,46 @@
...
+},
+"dependencies": {
+  "near-api-js": "^0.29.0",
+  "regenerator-runtime": "^0.13.5",
+  "vue": "^2.6.11"
+},
+  "jest": {
+    "testEnvironment": "near-cli/test_environment",
+    "testPathIgnorePatterns": [
+      "<rootDir>/assembly/",
```
chadoh, line 42: We shouldn't need this line in a Rust project. Also, are any of these Jest configs or dependencies being used? I don't see any tests! Please add tests, and if you don't want to use Jest with them, you can get rid of all this Jest stuff.

templates/vue/copy-dev-account.js
```diff
@@ -0,0 +1,7 @@
+let fs=require('fs')
```
chadoh, line 1: You can make this file executable (`chmod +x copy-dev-account.js`) if you add one line to the top:

    #!/usr/bin/env node

Then you could drop the `.js` from the filename and invoke it with `./copy-dev-account` instead of `node copy-dev-account`

Up to you. I like this pattern 🙂

templates/vue/src/App.vue
luciotato: indentation error, fixed, now in src/components/SignedIn.vue

templates/vue/src/main.js
luciotato: fixed

templates/vue/src/__mocks__/fileMock.js
luciotato: deleted dir __mocks__

templates/vue/packagejsons/rust/package.json
luciotato: Removed that, and reorganized the Vue App including canonical vue test units.

templates/vue/src/components/Notification.vue
luciotato: fixed

templates/vue/asconfig.js
luciotato: ok, I've upgraded to near-sdk-as@1.0.x

common/frontend/config.js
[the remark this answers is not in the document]
luciotato: Let's close this vue-related issue first, and then I can help you with that.

Maybe we can generate a file in /src/generated/config.js (.gitignoring src/generated) as part of the common build process after `near dev-deploy` 
The file will only export a single object with the selected configuration (based on process.env.NODE_ENV) including the contract name from `/neardev/dev-account`
This way you can delete config.js from all templates and also only the selected config data gets baked into the web app

templates/vue3/src/App.vue
luciotato: Let's close this issue first as defined. 

After that maybe we can open a new issue to enhance notifications in all templates and I can help with that. 

There's no "easy" way, the `<aside>` is using position:fixed, gotta make that adjustable. In the vanilla template we have to clone & insert into the DOM a new `<aside>` for each notification adjusting position & animation. For Vue we can use https://github.com/euvl/vue-notification and I guess there are similar solutions for React.

common/frontend/config.js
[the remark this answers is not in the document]
chadoh: I don’t like the idea of React apps or vanilla JS apps being generated with a `VUE_APP_CONTRACT_NAME` variable in them, and I think we need to fix that before we merge this. I’m totally ok adding a little duplication for now, and then we can totally remove the need for this file by putting these defaults in near-api-js

templates/vue3/src/App.vue
chadoh: Yeah, not sure I’d want the notification to be “more real” – it might be fine as-is? Regardless, I agree that it shouldn’t be fixed as part of this PR.

templates/vue/copy-dev-account.js
chadoh: Not insisting, I like your reasoning 😄

common/frontend/config.js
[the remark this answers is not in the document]
chadoh: Unfortunately that won’t be enough. `config.js` will need to be moved into each template dir and deleted from `common/frontend` entirely. The `index.js` script does not copy `config.js` from the template to avoid copying symlinks, and then it copies over the stuff in `common/frontend`. So the Vue template will end up with an incorrect config.js unless we go all the way with this. 

Seems like the Angular template needed this too (https://github.com/near/bounties/issues/24)

I’ll make a PR to add this to `master` separate from this PR, and once it gets merged you can rebase/merge it in. That should make your work easier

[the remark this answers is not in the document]
chadoh: I created #457, once the tests pass we can get it merged and then you can rebase/merge on latest `master`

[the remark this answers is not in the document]
chadoh: Merged #457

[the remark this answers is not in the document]
luciotato: looks like this PR from my branch: add-vue-template is still good to go, no need to rebase/merge

luciotato:
It's all green!
@chadoh could you merge this PR, please?

templates/vue/packagejsons/assemblyscript/package.json
```diff
@@ -0,0 +1,72 @@
...
+    "@vue/cli-plugin-unit-jest": "^4.5.4",
+    "@vue/cli-service": "~4.5.0",
+    "@vue/test-utils": "^1.0.3",
+    "env-cmd": "^10.1.0",
+    "eslint": "^6.7.2",
+    "eslint-plugin-vue": "^6.2.2",
+    "gh-pages": "^3.0.0",
+    "vue-template-compiler": "^2.6.11"
+  },
+  "eslintConfig": {
```
chadoh, line 42: Do we need this same config in the rust package.json?

templates/vue/packagejsons/assemblyscript/package.json
```diff
@@ -0,0 +1,72 @@
...
+    "build": "npm run build:contract && npm run build:web",
+    "build:contract": "asb",
+    "build:contract:debug": "asb --target debug",
+    "build:web": "node copy-dev-account.js && vue-cli-service build",
+    "dev:deploy:contract": "near dev-deploy",
+    "deploy:contract": "near deploy",
+    "deploy:pages": "gh-pages -d dist/",
+    "deploy": "npm run build && npm run deploy:contract && npm run deploy:pages",
+    "prestart": "npm run build:contract:debug && npm run dev:deploy:contract",
+    "start": "echo The app is starting! It will automatically open in your browser when ready && npm run serve",
```
chadoh, line 15: This message is not true with the way this is currently written. Is there an option we can pass to `vue-cli-service serve` to automatically open the app?

chadoh: Looks like you can pass `vue-cli-service --open`, please do that in both package.json files

common/frontend/config.js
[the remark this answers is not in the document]
chadoh: note that without a rebase/merge, `common/frontend/config.js` is still on your branch, so local testing with `./index.js my-vue-app --frontend vue` does not use the correct `config.js`

tests still work because they set `CONTRACT_NAME`, not `VUE_APP_CONTRACT_NAME`. Will they still pass once they use the correct `config.js`?

templates/vue/src/App.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+  <div id="root">
+      <SignedOut v-show="!isSignedIn" />
+      <SignedIn v-show="isSignedIn" />
```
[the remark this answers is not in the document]
luciotato: Yeah, you're right, the indentation for vue <templates> is configured in a separate `.eslintrc` rule. Configured & fixed.

common/frontend/config.js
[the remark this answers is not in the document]
luciotato: I removed `common/frontend/config.js` when you asked me to create separate configs. I'll merge anyway to be safe.

templates/vue/packagejsons/assemblyscript/package.json
luciotato: Yes, you're right. I completely re-did `eslintConfig` and put it in both packages. I also added the needed eslint config to check indentation in vue templates

templates/vue/packagejsons/assemblyscript/package.json
luciotato: Yes. Done.

luciotato: Done.

templates/vue/src/utils.js
```diff
@@ -0,0 +1,71 @@
+import { connect, Contract, keyStores, WalletConnection } from 'near-api-js'
```
chadoh, line 1: This `utils.js` file is not needed, is it? It should be a symlink I think

templates/vue/tests/unit/Notification.spec.js
```diff
@@ -0,0 +1,26 @@
+import { shallowMount } from '@vue/test-utils'
+import Notification from '@/components/Notification.vue'
```
chadoh, line 2: what is this `from '@/...'` notation?

templates/vue/src/utils.js
luciotato: If you soft-symlink that file you'll get a vue compiler error:
```
> cd templates/vue/src
> mv utils.js NOSYMLINK-utils.js
> ln -s ../../../common/frontend/utils.js
> cd ..
> npm run serve

ERROR  Failed to compile with 1 errors                                                                                                                                                                                            
This dependency was not found:
* config in /home/lucio/repos/create-near-app/common/frontend/utils.js
```
Since config is now local to the template and `utils.js` does `import getConfig from './config'`, the compiler tries to look for 'config.js' in the same dir as 'utils.js', that's it: '/common/frontend/'

One solution is to *hard* symlink `utils.js`, but that will mess up git. 
The second solution is to have a copy of both files in the template: utils.js and config.js

templates/vue/tests/unit/Notification.spec.js
luciotato: it means: "*from the root of the project*", but in this case it's not needed. I'll remove it.

chadoh: Whoa cool! Thanks for teaching me this!

templates/vue/src/utils.js
chadoh: Interesting! Parcel has no problem with this in the `templates/vanilla` or `templates/react` directories when I test on my machine.

The duplication of this file makes me wonder if we should stop putting it in `common/frontend`, as we did with `config.js` 🤔

luciotato: I agree with you, we should stop putting utils.js in common/frontend

Quangtran197:
1

PR #460, opened 2020-08-23
feat(templates): add angular template

faragly:


templates/react/packagejsons/rust/package.json
```diff
@@ -12,7 +12,7 @@
     "deploy": "npm run build && npm run deploy:contract && npm run deploy:pages",
     "prestart": "npm run build:contract && npm run dev:deploy:contract",
     "start": "echo The app is starting! It will automatically open in your browser when ready && env-cmd -f ./neardev/dev-account.env parcel src/index.html --open",
-    "dev": "nodemon --watch assembly -e ts --exec \"npm run start\"",
+    "dev": "nodemon --watch contract -e ts --exec \"npm run start\"",
```
chadoh, line 15: Thank you!!

templates/angular/src/app/app.component.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    it('should display greeting text', () => {
+      const compiled = fixture.nativeElement
+      const h1 = compiled.querySelector('main h1')
+
+      expect(h1.textContent).toContain('Hello test.near')
+    })
+
+    it('should call the `logout` method on `Sign out` action', () => {
+      const btn = fixture.debugElement.query(By.css('button.link'))
```
chadoh, line 94: In the interest of setting up a project for success at it grows larger, could this button be found by using its innerHTML, rather than a CSS selector? I would imagine that an app could have lots of `button.link` elements as it grows

templates/angular/src/app/app.component.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+    app = fixture.componentInstance
+    spyWindow = TestBed.inject(WINDOW) as jasmine.SpyObj<Window>
+    fixture.detectChanges()
+  })
+
+  it('should create the app', () => {
+    expect(app).toBeTruthy()
+  })
+
+  describe('signedIn', () => {
```
chadoh, line 61: The `describe('signedIn'` and `describe('signedOut'` seem switched

chadoh (changes requested):
I haven't used Angular since 2015, and had no idea how it changed since then. This PR was a great way to learn about it! I found your code very straightforward (Angular itself, on the other hand, I still find a bit awkward!)

See a few small nits, questions, and ideas for improvement below. None of them should take long to address!

templates/angular/set-contract-name.js
```diff
@@ -0,0 +1,16 @@
+const { writeFileSync } = require ('fs')
+require('dotenv').config({ path: './neardev/dev-account.env' })
+const colors = require('colors')
+
+const targetPath = './src/contract-name.js'
+const contractNameFileContent = `module.exports = '${process.env.CONTRACT_NAME || 'near-blank-project'}'`
```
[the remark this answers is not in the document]
faragly: We cannot access the environment variables in angular. If we try to get the value of `process.env.CONTRACT_NAME`, then instead of the name of the contract we will get `undefined`.

https://stackoverflow.com/questions/49496438/use-process-env-in-angular-5-environment#:~:text=the%20Angular%20code.-,process.,truly%20needs%20to%20be%20dynamic.

To create similar variables for different build modes, I could use `src/environments/environment.ts`, but this is an unnecessary complication, since the developer will supplement this file with his own variables, and `set-contract-name.ts` will need to pointwise change the value of the `contractName` key (using `replace-in-files`).

templates/angular/extra-webpack.config.ts
```diff
@@ -0,0 +1,7 @@
+import * as webpack from 'webpack';
+
+export default {
+  node: {
+    fs: 'empty'
```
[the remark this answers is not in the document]
faragly: image
Without this options, we will get errors like in the screenshot above. This configuration returns an empty object for the `fs` module, we don't need it inside the Angular application.
https://webpack.js.org/configuration/node/

templates/angular/src/app/app.component.scss
```diff
@@ -0,0 +1,35 @@
+button.link {
+  float: right;
```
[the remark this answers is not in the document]
faragly: > This seems wrong – why should every button with a link class be floated right?

Styles are not inherited by any components nested within the template - https://angular.io/guide/component-styles. These styles is not global, any other `button.link` will not inherit the styles specified in this file.

> Why is this file needed at all? > Does Angular not support inline styles?
I removed this component styles file and made inline styles in component template.

> If you're able to put these styles back inline, would that also mean the scss library import is not needed? I'd rather keep things as unopinionated as possible, though I understand that Angular has a lot of opinions built in, so maybe removing scss-related dependencies is not useful in this case.
You're right, removing scss-related dependencies is not useful in this case. Angular has built-in support all major CSS preprocessors: `sass/scss`, `less`, and `stylus`. We can use any of them, or even several (I don't know why a developer might need this, but still).

templates/angular/src/app/app.component.spec.ts
faragly: Fixed.

index.js
```diff
@@ -88,16 +88,24 @@ const createProject = async function({ contract, frontend, projectDir, veryVerbo
...
-    await replaceInFiles({ files: `${projectDir}/src/*`, from: /accountId:/g, to: 'account_id:' })
+    await replaceInFiles({ files: `${projectDir}/src/**/*`, from: /getGreeting/g, to: 'get_greeting' })
+    await replaceInFiles({ files: `${projectDir}/src/**/*`, from: /setGreeting/g, to: 'set_greeting' })
+    await replaceInFiles({ files: `${projectDir}/src/**/*`, from: /assembly\/main.ts/g, to: 'contract/src/lib.rs' })
+
+    if (frontend === 'angular') {
+      await replaceInFiles({ files: `${projectDir}/src/**/*`, from: /{ accountId:/g, to: '{ account_id:' })
+    } else {
+      await replaceInFiles({ files: `${projectDir}/src/*`, from: /accountId:/g, to: 'account_id:' })
+    }
```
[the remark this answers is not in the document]
faragly: Removed the `if` statement, thanks!

templates/angular/src/app/app.component.spec.ts
faragly: Fixed.

templates/angular/set-contract-name.js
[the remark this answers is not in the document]
chadoh: Gotcha, that's unfortunate. For my own understanding, is this `set-contract-name` a common workaround or your own invention?

templates/angular/src/app/app.component.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+  describe('signedOut', () => {
+    it('should display welcome message', () => {
+      const compiled = fixture.nativeElement
+      const h1 = compiled.querySelector('main h1')
+
+      expect(h1.textContent).toContain('Welcome to NEAR!')
+    })
+
+    it('should call the `login` method on `Sign in` action', () => {
+      const btn = fixture.debugElement.query(By.css('button'))
```
chadoh, line 70: Copying my old message here:

> In the interest of setting up a project for success at it grows larger, could this button be found by using its innerHTML, rather than a CSS selector? I would imagine that an app could have lots of button.link elements as it grows

Selecting any `button` seems like it might run into problems even more quickly, as people add new buttons to their apps. Can you select by `innerHTML`?

templates/angular/set-contract-name.js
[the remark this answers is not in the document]
faragly: This solution is based on [the article](https://medium.com/javascript-in-plain-english/setup-dotenv-to-access-environment-variables-in-angular-9-f06c6ffb86c0), but the target file is different. This is done so that the developer can still work with `environment*.ts` files, set his own variables, and his changes will not be overwritten after `set-contract-name` processing.

templates/angular/src/app/app.component.spec.ts
faragly: I fixed the selector for the `Sign in` button.

chadoh (approved):
Thanks so much, @faragly!

templates/angular/src/contract-name.js
```diff
@@ -0,0 +1 @@
+module.exports = ''
```
chadoh, line 1: @faragly does this file need to be included in the source? It seems like a build artifact that should be gitignored, since it will change based on whether you're using `yarn dev` or `yarn start`, and could also change between different runs of `yarn dev` if you clear your `neardev/dev-account` file

chadoh: See https://github.com/near/create-near-app/pull/510

faragly: You are right, this file changes frequently and should be gitignored. But if we delete this file, then an error occurs when importing a non-existent file (these errors are now visible in Travis). I have not yet found a concise way to get rid of this file.

PR #464, opened 2020-08-25
refactor(utils.js): mv out of common; rm onSubmit

chadoh:
onSubmit never felt quite right as a shared function, given that it manipulates the DOM

Additionally, the vue cli added in #452 does not work with `utils.js` being in a symlink and `config.js` being in the project directory, so they both need to live in the project directory.

Ideally, near-api-js could be simplified to a point that the helpers in utils.js are superfluous. That will be a better way to get rid of the duplication in each template, rather than using a symlink.


templates/react/src/App.js
```diff
@@ -83,12 +83,34 @@ export default function App() {
...
+              'Maybe you need to sign out and back in? ' +
+              'Check your browser console for more info.'
+            )
+            throw e
+          } finally {
+            // re-enable the form, whether the call succeeded or failed
+            fieldset.disabled = false
+          }
 
           // update local `greeting` variable to match persisted value
```
luciotato, line 115: I would move all this code inside the try-catch, so if there's an error, no notification other than the alert is raised

chadoh: the code in question does two things

1. update local `greeting` variable to match persisted value
2. show notification, then hide again after 10 seconds

We only want to do these things if the code in the `try` block completes _successfully_. As-written, this is exactly what it will do. 

We could put it in the `try` block below the `contract.setGreeting` call. However, this seems less straightforward to me – the `catch` & `finally` are there to deal with the `contract.setGreeting` call failing/completing. If there were several more steps in the `try` block, it would be unclear which part of it might throw the error we want to deal with.

luciotato: > If there were several more steps in the try block, it would be unclear which part of it might throw the error we want to deal with

ok, yes, you're right.

PR #502, opened 2020-10-06
feat: add overflow-checks to Rust release profile

### LINKED ISSUE #498: Oveflow and paranoid checks should be enabled by default for Rust contracts
Rust supports overflow checks for integer arithmetics, however they are disabled by default for optimised builds.

[These checks are important because of an attack vector when dealing with user balances](https://ethereum.stackexchange.com/questions/8551/security-review-checklist-for-a-smart-contract).

Checks should be enabled in the generated `Cargo.toml` by default

https://github.com/near/create-near-app/blob/master/common/contracts/rust/Cargo.toml

More information here:

https://stackoverflow.com/q/64129432/315168

chadoh:
Fixes #498 

By default, Rust has overflow checks enabled for debug builds, but disabled in optimized release builds. This adds the checks for production builds as well.

### DIFF
--- a/common/contracts/rust/Cargo.toml
+++ b/common/contracts/rust/Cargo.toml
@@ -17,6 +17,8 @@ opt-level = "z"
 lto = true
 debug = false
 panic = "abort"
+# Opt into extra safety checks on arithmetic operations https://stackoverflow.com/a/64136471/249801
+overflow-checks = true
 
 [workspace]
 members = []


common/contracts/rust/Cargo.toml
```diff
@@ -17,6 +17,8 @@ opt-level = "z"
 lto = true
 debug = false
 panic = "abort"
+# Ensure safe uint handling https://stackoverflow.com/a/64136471/249801
```
frol, line 20: This comment sounds scary to me. How about: "Opt into extra safety checks on arithmetic operations"?

frol (approved):

PR #509, opened 2020-10-13
refactor: find styles when rapid-developing Angular

chadoh:
If you `cd templates/angular`, `yarn`, then `yarn dev`, the styles didn't load without this, because Angular by default doesn't follow symlinks


chadoh:
@faragly does this seem sensible to you?

faragly:
@chadoh Unfortunately, I didn't start the project this way, `preserveSymlinks: true` in the `architect.build.options` object solve the problem.

PR #510, opened 2020-10-13
refactor(angular): gitignore contract-name

chadoh:
This file is an auto-generated build artifact which can change frequently throughout the development cycle

### DIFF
--- a/templates/angular/.gitignore
+++ b/templates/angular/.gitignore
@@ -1,0 +1,1 @@
+src/contract-name.js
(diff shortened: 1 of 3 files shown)


chadoh:
Copying a [comment](https://github.com/near/create-near-app/pull/460#discussion_r506937778) from @faragly here for better visibility:

> You are right, this file changes frequently and should be gitignored. But if we delete this file, then an error occurs when importing a non-existent file (these errors are now visible in Travis). I have not yet found a concise way to get rid of this file.

gagdiez:
angular was removed

PR #517, opened 2020-10-15
build: test in diff folder, avoid false positives

chadoh:
I noticed that sometimes the tests could pass, even though generating a
particular contract+frontend combo failed. This could happen when the
frontend had dependencies covered by the root `package.json`, but not
actually included its own `package.json`

Why did this happen? Because we were creating the test project inside
the root folder, so when the test project tried to look up a given
dependency, it was able to find it in the root project's node_modules
folder.

We need to make sure we test in a realistic environment, where the
project cannot depend on the root folder's node_modules

### DIFF
--- a/templates/angular/packagejsons/rust/package.json
+++ b/templates/angular/packagejsons/rust/package.json
@@ -55,6 +55,7 @@
     "nodemon": "^2.0.4",
     "protractor": "~7.0.0",
     "puppeteer": "^5.3.0",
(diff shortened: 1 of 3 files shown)


chadoh:
Note that this revealed that a fresh Angular project with Rust did not work, since it did not have `shelljs` installed

I also added `shelljs` as a dependency to Vue, even though it seemed to be an implicit dependency already, as I think it's better to make it explicit

PR #574, opened 2020-12-03
fix: simplify compile; rm test script; fix 'yarn dev'

### LINKED ISSUE #573: `compile` script doesn't propagate build errors
from @vgrichina 

> looks like https://github.com/near/create-near-app/blob/master/common/contracts/assemblyscript/compile
> doesn't propagate error code from `asb`
> so when you create project using `create-near-app` and do `yarn dev` – it'll redeploy old version instead of stopping on error
> 
> seems like there was related bug https://github.com/shelljs/shelljs/issues/735
> 
> seems like bug is still in shelljs 0.8.4 🤷‍♂️

chadoh:
Fixes #573 

We learned some things recently:

* Executable NodeJS files are less familiar to the average user than using `node somefile.js`
* If we're going to put a script file in the generated project, better to have it be simple and commented voluminously than have it give useful help output when run with `node some-script --help`
* The `verbose` mode was gratuitous
* Installing assemblyscript dependencies every time you compile is excessive. Instead, we can install them once, right after the project is generated. Future updates to `contract/package.json` will be kicked off by the new project's owner, who will presumably also use `yarn install` in tandem with such an update.

Additionally, there were some bugs or inevitable future bugs in the code:

* `compile` script wasn't propagating build errors (if build command failed, `compile` script still exited with status `0`, meaning it passed and swallowed the build command's error)
* hard-coding `greeter` package name means that as soon as people change their package to be called something else, the symlinking logic breaks

And two more things:

* Remove `test` script, since extra executable files are less familiar to beginners than some extra verbosity in the package.json "test" script. This means that the main `create-near-app` command in `index.js` needs to replace `npm run test` with `cargo test` for Rust projects.
* Fix the `yarn dev` command to set up `nodemon` to watch `ts`
files properly (and `rs` files for Rust projects), so that the contract
will automatically be re-compiled & deployed any time it's changed.


thisisjoshford (approved):
Love the comments! Much easier to parse out now :) Still would love the elimiination of "compile file" when [PR #603](https://github.com/near/near-cli/pull/603) is merged.

index.js
```diff
@@ -120,6 +122,9 @@ const createProject = async function({ contract, frontend, projectDir, veryVerbo
   if (hasNpm || hasYarn) {
     console.log('Installing project dependencies...')
     spawn.sync(hasYarn ? 'yarn' : 'npm', ['install'], { cwd: projectDir, stdio: 'inherit' })
+    if (contract === 'assemblyscript') {
+      spawn.sync('npm', ['install', '--legacy-peer-deps'], { cwd: `${projectDir}/contract`, stdio: 'inherit' })
```
vgrichina, line 126: @chadoh why not using `yarn` here when available similar to command above?

common/contracts/assemblyscript/compile.js
```diff
@@ -0,0 +1,51 @@
...
+// Use the correct build command based on the `--debug` flag
+const buildCmd = debug
+  ? 'npm run build:debug'
+  : 'npm run build'
+
+// Execute the build command, storing exit code for later use
+const { code } = sh.exec(buildCmd)
+
+// Assuming this is compiled from the root project directory, link the compiled
+// contract to the `out` folder –
```
vgrichina, line 36: @chadoh note that it is possible to pass `--wasmFile` argument to these commands if necessary to deploy from any file

index.js
chadoh: I was thinking we can start doing that once the `--legacy-peer-deps` is [no longer needed](https://github.com/near/near-sdk-as/issues/312), just to save a little bit of code here. Probably wasn't worth it, but ¯\_(ツ)_/¯

common/contracts/assemblyscript/compile.js
chadoh: This isn't new. near-api-js always used to copy/compile to `ROOT/out/main.wasm`. What's newer is symlinking instead. And once https://github.com/near/near-cli/pull/603 is merged, we can get rid of the symlinking step altogether and remove the `compile` script for AssemblyScript.

PR #835, opened 2021-03-25
fix(generated): fixes #831: copy-update instead of linking

### LINKED ISSUE #831: create-near-app vanilla on windows 10 not working
created a vanilla app
everything gets created but `yarn start` fails with an error

To reproduce: (windows 10)
>create-near-app vanilla
>cd vanilla
>yarn start
image

luciotato:
use `cp -u` instead of `sh.ln`.  `sh.ln` fails on windows 10

### DIFF
--- a/templates/angular/package.json
+++ b/templates/angular/package.json
@@ -55,7 +55,7 @@
     "near-cli": "~1.6.0",
     "nodemon": "~2.0.4",
     "protractor": "~7.0.0",
-    "puppeteer": "~5.5.0",
+    "puppeteer": "~9.0.0",
     "shelljs": "~0.8.4",
     "tslint": "~6.1.0",
     "typescript": "~4.0.5"


vgrichina (approved):
lgtm, but TBH I don't know why we need this script instead of:
1) Running `npm run build` or `npm run build:debug` directly
2) Referencing the specific file path in deploy instead of copying files

Check out how build is setup in https://github.com/vgrichina/bananaswap/blob/16a9e72c35b8f17ce8fb84034113e5d0b39c9519/package.json#L7

cc @willemneal @chadoh maybe you have better context on why this script is needed

chadoh:
We have `compile.js` scripts for both Rust and AssemblyScript mostly because we currently use this "copy/symlink to `/out/main.wasm`" approach, and the copying logic is complex. If we didn't need to copy the wasm file, we could get rid of the `compile.js` script and use the `npm run build` commands directly in the root package.json as you suggest. 

Note that we need create-near-app to work correctly for both AssemblyScript and Rust. The `compile.js` scripts are quite similar:

diff common/contracts/*/compile.js


Note that we would not be able to _only_ have the `npm run` commands, but would need to use `replaceInFiles` to put in the Rust versions if that's the contract language selected.

Also, as soon as https://github.com/near/near-cli/pull/603 gets merged, I would love to get rid of these compile.js scripts and use Vlad's suggested approach. @luciotato is that something you would want to help with?

common/contracts/assemblyscript/compile.js
```diff
@@ -43,8 +43,10 @@ if (code === 0 && calledFromDir !== __dirname) {
   const outFile = `./build/${debug ? 'debug' : 'release'}/${packageName}.wasm`
   sh.mkdir('-p', linkDir)
   sh.rm('-f', link)
-  const linkPath = path.relative(linkDir, outFile)
-  sh.ln('-s', linkPath, link)
+  //fix #831: copy-update instead of linking .- sometimes sh.ln does not work on Windows
+  //const linkPath = path.relative(linkDir, outFile)
+  //sh.ln('-s', linkPath, link)
```
chadoh, line 48: Can get rid of these comments

chadoh (changes requested):
Need the same changes in the `rust/compile.js` script

chadoh:
Note that Travis is not currently running tests for this project. Do we know this works for macOS & Linux?

chadoh:
oh, I think you'll want to put the "fixes #831" in the PR description so that #831 gets auto-closed when this is merged

common/contracts/assemblyscript/compile.js
luciotato: leave this 

`//fix #831: use copy-update instead of linking .- sometimes sh.ln does not work on Windows`

so this is not reintroduced in the future

luciotato: #fix,  #fixes, #fixed... all link an Issue to a PR

volovyk-s:
@luciotato is this PR still relevant? If so, please, merge the current muster. We have fixed the CI.

luciotato:
ok, I didn't see @chadoh request. I've added the changes as requested

volovyk-s:
@luciotato please, rebase. We have a new CI in master.

luciotato:
rebased. now is https://github.com/near/create-near-app/pull/913

PR #840, opened 2021-03-29
HashMap replaced with LookupMap

volovyk-s:
We should use `near_sdk::collections` instead of `std::collections` in our examples.

### DIFF
--- a/common/contracts/rust/src/lib.rs
+++ b/common/contracts/rust/src/lib.rs
@@ -15,7 +15,7 @@
 use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
 use near_sdk::wee_alloc;
 use near_sdk::{env, near_bindgen};
-use std::collections::HashMap;
+use near_sdk::collections::LookupMap;
 
 #[global_allocator]
 static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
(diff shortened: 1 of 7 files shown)


common/contracts/rust/src/lib.rs
```diff
@@ -15,17 +15,23 @@
...
+use near_sdk::collections::LookupMap;
 
 #[global_allocator]
 static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
 
 // Structs in Rust are similar to other languages, and may include impl keyword as shown below
 // Note: the names of the structs are not important when calling the smart contract, but the function names are
 #[near_bindgen]
-#[derive(Default, BorshDeserialize, BorshSerialize)]
+#[derive(BorshDeserialize, BorshSerialize)]
```
mikedotexe, line 26: Let's try using this brand new thing: `PanicOnDefault` like here:
https://github.com/near/near-sdk-rs/blob/master/examples/fungible-token/ft/src/lib.rs#L30

I believe you can then remove the `impl Default for Welcome…` block.

vgrichina (approved):

vgrichina:
@volovyk-s looks like commit message needs to follow https://github.com/near/create-near-app#about-commit-messages to unbreak CI

PR #846, opened 2021-03-30
eslint GitHub Action added

volovyk-s:
Let's start fixing CI here step by step.
It's a simple linter build with GitHub Actions.


chadoh (changes requested):
I'm fine with adding superlinter IF we also configure the project so that our editors can easily find the same configuration it uses.

I want to see:

* relevant superlinter setup in the project itself, so that editors can give the same warnings as CI
* a `lint` script added to root `package.json` so that running `yarn lint` locally matches what the CI job tests
* all superlinter-discovered errors addressed so this new CI job passes

vgrichina:
> * a `lint` script added to root `package.json` so that running `yarn lint` locally matches what the CI job tests
> * all superlinter-discovered errors addressed so this new CI job passes

agree 100% on these.

> * relevant superlinter setup in the project itself, so that editors can give the same warnings as CI


I don't think this one should be a blocker. It's ok if you have to run a script in `package.json`, unless IDE would have conflicting config showing different errors.

mikedotexe:
I just talked to Serhii about this and super linter isn't as robust as it could be. Yet.
As he looked into it we'd need to spin up a Docker just to check locally.
https://github.com/github/super-linter/blob/master/docs/run-linter-locally.md
I look forward to GitHub improving the tool, but until then let's just use `eslint` and not spend much more time.

volovyk-s:
@chadoh @vgrichina @mikedotexe 
`Super Linter` changed to `eslint`

mikedotexe (approved):

PR #847, opened 2021-03-31
fix(generated): fix react git ignore

austinabell:
the `.cache` and `dist` folders aren't being ignored by default.

Also, for Rust contracts the `target` directory wasn't being ignored, but looking at the code it seems to be doing it correctly so it might just not be released? If not, happy to fix that, but it wasn't intuitive for me how to run from source (my JS is rusty sorry)

### DIFF
--- a/templates/react/near.gitignore
+++ b/templates/react/near.gitignore
@@ -4,11 +4,13 @@
 /node_modules
 /.pnp
 .pnp.js
+
+# build
 /out
+/dist
 
-#keys
+# keys
 /neardev
-
 
 # testing
 /coverage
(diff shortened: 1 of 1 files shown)


vgrichina:
@austinabell thanks!

looks like needs to be updated to pass CI:
https://travis-ci.com/github/near/create-near-app/builds/221825003

commit message should match expected format:
https://github.com/conventional-changelog/commitlint/#what-is-commitlint

austinabell:
> @austinabell thanks!
> 
> looks like needs to be updated to pass CI:
> https://travis-ci.com/github/near/create-near-app/builds/221825003
> 
> commit message should match expected format:
> https://github.com/conventional-changelog/commitlint/#what-is-commitlint

Force pushed with new commit. Let me know if the new commit is in the right form

PR #917, opened 2021-04-26
Add mixpanel tracking for type of frontend and contract

icerove:
Add mixpanel tracking for type of frontend and contract user choose

### DIFF
--- a/index.js
+++ b/index.js
@@ -56,6 +56,7 @@ function copyDir (source, dest, { skip, veryVerbose } = {}) {
 const createProject = async function({ contract, frontend, projectDir, veryVerbose }) {
   const templateDir = `/templates/${frontend}`
   const sourceTemplateDir = __dirname + templateDir
+  mixpanel.track(frontend, contract)
 
   console.log(`Copying files to new project directory (${projectDir}) from template source (${sourceTemplateDir}).`)
 
@@ -165,8 +166,6 @@ We suggest that you begin by typing:`)
     {bold cd ${projectDir}}
     {bold ${runCommand} dev}`)
   }
-
-  mixpanel.track(frontend, contract)
 
   console.log(chalk`
 Happy hacking!`)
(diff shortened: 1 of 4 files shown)


tracking.js
```diff
@@ -0,0 +1,26 @@
+const MIXPANEL_TOKEN = 'df164f13212cbb0dfdae991da60e87f2'
+
+const chalk = require('chalk')
+const mixpanel = require('mixpanel').init(MIXPANEL_TOKEN);
+
+const track = async (frontendType, contractType) => {
+    try {
+        console.log(chalk`
+NEAR collect the type of UI and contract to improve developer experience. But NEAR will never send private information.
```
volovyk-s, line 9: I'm not sure about the grammar here. @TiffanyGYJ can you please take a look?

index.js
```diff
@@ -165,6 +166,8 @@ We suggest that you begin by typing:`)
     {bold ${runCommand} dev}`)
   }
 
+  mixpanel.track(frontend, contract)
```
volovyk-s, line 169: @icerove I would not put `NEAR collect the type of UI and contract...` info at the end of the build. We really want developers to concentrate on `cd appName` and `yarn dev` commands. I would rather put it at the beginning of the function.

volovyk-s (changes requested):
@icerove Looks good overall.

tracking.js
TiffanyGYJ: I would write something like this: 

NEAR collects basic information on the type of commands used to improve developer experiences. All data are anonymous, therefore, no personal information that could identify you will be shared.

TiffanyGYJ:
I would write something like this:

NEAR collects basic information on the type of commands used to improve developer experiences. All data are anonymous, therefore, no personal information that could identify you will be shared.

volovyk-s (approved):

PR #928, opened 2021-05-03
add checking filter tests 

icerove:
add checking filter tests, only tracking for real users


volovyk-s:
Sorry @icerove , haven't seen this PR. The fix is already merged.

PR #958, opened 2021-05-11
v2.0.0

### LINKED ISSUE #831: create-near-app vanilla on windows 10 not working

volovyk-s:
### Rust autosetup added

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "create-near-app",
-  "version": "1.3.6",
+  "version": "2.0.0",
   "description": "generate new near blank project with different type",
   "main": "index.js",
   "files": [


volovyk-s:
> why 2.0? Are we shipping some major breaking changes?
> 
> also generally bumping version yourself conflicts with `np` release process

No major updates, but there is a breaking change. Now we need to accept `Rustup` installation during the project creation (if it's a rust project). So according to https://semver.org/ it should be a major release.

GitHub was not able to process PR from `np`. There was an error, something like `url is too big, can not openen`.

PR #1153, opened 2021-07-23
feat: Update rust contract to 4.0 pre release

austinabell:
self-explanatory, I made a few other related changes to make it easier to read and get started.

Also tested manually as well as running tests to make sure everything was working correctly.

built contract size before was `112K` and is now `92K`

### DIFF
--- a/common/contracts/rust/Cargo.toml
+++ b/common/contracts/rust/Cargo.toml
@@ -5,10 +5,10 @@ authors = ["Near Inc <hello@near.org>"]
 edition = "2018"
 
 [lib]
-crate-type = ["cdylib", "rlib"]
+crate-type = ["cdylib"]
 
 [dependencies]
-near-sdk = "3.1.0"
+near-sdk = "4.0.0-pre.1"
 
 [profile.release]
 codegen-units = 1
--- a/common/contracts/rust/Cargo.lock
+++ b/common/contracts/rust/Cargo.lock
@@ -333,9 +333,9 @@ dependencies = [
 
 [[package]]
 name = "near-sdk"
-version = "3.1.0"
+version = "4.0.0-pre.1"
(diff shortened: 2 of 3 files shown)


common/contracts/rust/Cargo.toml
```diff
@@ -5,10 +5,10 @@ authors = ["Near Inc <hello@near.org>"]
 edition = "2018"
 
 [lib]
-crate-type = ["cdylib", "rlib"]
+crate-type = ["cdylib"]
 
 [dependencies]
-near-sdk = "3.1.0"
+near-sdk = "4.0.0-pre.1"
```
volovyk-s, line 11: Should we wait for the release of 4.0.0?

austinabell: @mikedotexe maybe you have an opinion here. This release is intended to be stable, without any core fundamental changes. Figured it would be a good way to get users on this version before we release the stable version, and it allows users to have the most up-to-date APIs and without extra bloat.

mikedotexe: Okay, thanks for clarifying. Yes, I'm down to use prereleases like this. We haven't done this before, but I'm on board. Previously we've been using prereleases as a kind of hack to (if I recall) make sure CI passes across the board with different projects using it.

Do you have concerns, @volovyk-s ?

volovyk-s: After the last clarification from @austinabell, I have no concerns anymore.

volovyk-s (approved):

common/contracts/rust/Cargo.toml
austinabell: But, just for posterity, I do believe we should avoid updating this to pre-releases in the future, but in this case, there are a few rough edges on the API we want to move away from ASAP before users get used to it.

PR #1451, opened 2021-11-23
add js frontend

starpav:


volovyk-s:
@starpav we are trying to keep this project lightweight. It should not have any `complex` projects as NFTs, etc. Also, it will be replaced with more comprehensive examples in the future.

PR #1734, opened 2022-04-27
updated react version and implementation

doriancrutcher:
I've updated the react template to the latest version of react and changed the implementation so that react doesn't show an error anymore to the users.


volovyk-s:
@doriancrutcher cool! Can you please fix the linter errors?

PR #1786, opened 2022-05-17
Update examples and simplify code

gagdiez:
All the examples were updated in the following way:
 - the contract was simplified (only stores a general message)
 - the rust version uses the latest version
 - the package.jsons have been updated

I removed all the versions except "React" and "Vanilla", since we
only get requests for those two.

In my local computer all tests pass, and the frontend works perfectly.


gagdiez:
@thisisjoshford Created a PR to simplify create-near-app, so we can use it in our new docs.

index.js
```diff
@@ -177,17 +160,17 @@ Happy hacking!`)
...
     desc: 'template to use',
-    choices: ['vanilla', 'react', 'vue', 'angular'],
+    choices: ['vanilla', 'react'],
     default: 'vanilla',
   })
   .option('contract', {
     desc: 'language for smart contract',
     choices: ['assemblyscript', 'rust'],
-    default: 'assemblyscript'
+    default: 'rust'
```
volovyk-s, line 173: Was it discussed with the team? cc @thisisjoshford

volovyk-s (commented):
Nice work! It's cool that we will have integration with Workspaces as a default.
- I'm not sure why do we need to simplify the status contract. It is simple already. And it is not a problem to delete extra code. (not a blocker)
- We should fix CI before merging (blocker)

gagdiez:
@volovyk-s CI is fixed. We decided to simplify the example so people do not need to worry about "predecessors" the first time they see a smart contract.

volovyk-s:
@gagdiez ok
But CI is still not showing the status for me:
image

gagdiez:
@volovyk-s I do not understand why it is not triggering actions here. In [my repo](https://github.com/gagdiez/create-near-app) actions are working, and all the tests pass. Any ideas why this could be happening and how to solve it?

volovyk-s:
@gagdiez  Oh, now I got it. You are not a part of GitHub NEAR org. Do you know if you can become one? (now you are only in near-examples).

gagdiez:
@volovyk-s will merge this into a branch, and from there we can see if the actions are triggered when merging to main

PR #1862, opened 2022-08-02
JS SDK support

maxhr:
- Major refactor
- JS SDK
- CLI user prompts


volovyk-s:
Since it will be the main way to create JS contract, let's add instructions here: https://github.com/near/near-sdk-js/pull/145

README.md
```diff
@@ -8,29 +8,29 @@ Quickly build apps backed by the [NEAR](https://near.org) blockchain
...
 
 Using [npm's npx](https://blog.npmjs.org/post/162869356040/introducing-npx-an-npm-package-runner):
 
-    npx create-near-app [options] new-awesome-project
+    npx create-near-app
 
-**Or**, if you prefer [yarn](https://classic.yarnpkg.com/en/docs/cli/create/):
+You can create contracts written in:
 
-    yarn create near-app [options] new-awesome-project
```
volovyk-s, line 26: How to specify the name of the project now? Should we mention it here?

README.md
```diff
@@ -8,29 +8,29 @@ Quickly build apps backed by the [NEAR](https://near.org) blockchain
...
+- [JavaScript](https://docs.near.org/develop/quickstart-guide)
+- AssemblyScript
 
-Without any options, this will create a project with a **vanilla JavaScript** frontend and an [AssemblyScript](https://docs.near.org/develop/quickstart-guide) smart contract
+You can create a frontend template in:
 
-Other options:
-
-* `--frontend=react` – use [React](https://reactjs.org/) for your frontend template
-* `--contract=rust` – use [Rust](https://docs.near.org/develop/prerequisites) for your smart contract
```
volovyk-s, line 33: Why are we deleting flags here?

src/app.ts
```diff
@@ -0,0 +1,99 @@
...
+    show.argsError();
+    return;
+  } else if (argsValid === 'ok') {
+    config = args as UserConfig;
+  }
+
+  show.welcome();
+
+  // Check node.js version
+  const current = process.version;
```
volovyk-s, line 32: Why we are not doing this is `package.json`?
```JS
  "engines" : { 
    "npm" : ">=7.0.0",
    "node" : ">=16.0.0"
  }
```

src/checks.ts
```diff
@@ -0,0 +1,30 @@
+import {CreateProjectParams} from './types';
+import chalk from 'chalk';
+
+export function preMessage(settings: CreateProjectParams) {
+  switch(settings.contract) {
+    case 'assemblyscript':
```
volovyk-s, line 6: Do we still need this AS-related code?

src/make.ts
```diff
@@ -0,0 +1,135 @@
...
+    verbose,
+    skip: skip.map(f => path.join(contractSourceDir, f))
+  });
+
+  // copy tests
+  let sourceTestDir = `${rootDir}/integration-tests`;
+  if (tests === 'workspaces') {
+    switch(contract) {
+      case 'js':
+      case 'assemblyscript':
```
volovyk-s, line 59: Are we deleting it completely?

volovyk-s (commented):
The fact that CI does not work for some people in our org is annoying. Why is this happening?

.github/workflows/test-matrix.yml
```diff
@@ -5,24 +5,16 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, macos-latest]
-        contract: [assemblyscript, rust]
-        frontend: ['vanilla', 'react', 'none']
```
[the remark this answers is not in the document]
maxhr: Tests are separate now to 2 workflows:
- test-matrix: tests generation of all templates and compares to their saved snapshots (see `test/make.test.ts`)
- test-e2e: generates templates, npm installs, and runs contract tests (`test/e2e.sh` - this works only on linux in github-actions, for some reason on macos it times-out

currently e2e break because of this: https://github.com/near/near-cli/issues/796
the CLI expects an input on first run

maxhr:
> The fact that CI does not work for some people in our org is annoying. Why is this happening?

@volovyk-s it still breaks on M1 because of workspaces-rs. Is that what you mean?

gagdiez:
I think @volovyk-s means that actions are not being triggered, and that's because the PR comes from another repo. Maybe we should migrate the branch here.

src/make.ts
volovyk-s: AS, duplicated conversation.

volovyk-s:
> > The fact that CI does not work for some people in our org is annoying. Why is this happening?
> 
> @volovyk-s it still breaks on M1 because of workspaces-rs. Is that what you mean?

I meant what @gagdiez said.
Regarding M1 - it should work, both Workspaces RS and JS support M1 on the latest version. But those versions are breaking Mac CI (known issue: https://github.com/near/nearcore/issues/7235).

package.json
```diff
@@ -1,22 +1,67 @@
...
+    "husky": "8.0.1",
+    "jest": "28.1.3",
+    "node-dir": "0.1.17",
+    "release-it": "15.0.0",
+    "ts-jest": "28.0.7",
+    "typescript": "4.7.4"
   },
   "repository": "https://github.com/near/create-near-app.git",
-  "author": "Yifang Ma <yifang@near.org>",
+  "author": "Near Inc",
```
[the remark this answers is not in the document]
idea404: Legally it is "Near Inc d/b/a Pagoda" which is effectively synonymous, so it's not wrong but I suppose the preferred name is Pagoda :)

test/make.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    ['assemblyscript', 'none', 'js'],
+    ['assemblyscript', 'react', 'rust'],
+    ['assemblyscript', 'vanilla', 'rust'],
+    ['assemblyscript', 'none', 'rust'],
+    ['rust', 'react', 'js'],
+    ['rust', 'react', 'rust'],
+    ['rust', 'vanilla', 'js'],
+    ['rust', 'vanilla', 'rust'],
+    ['rust', 'none', 'js'],
+    ['rust', 'none', 'rust'],
```
idea404, line 26: This looks like it will take some effort to maintain. Can't there be elsewhere a definition of supported SC & FE languages and create a matrix programmatically from that?

.github/workflows/test-e2e.yml
```diff
@@ -0,0 +1,26 @@
...
+  workflow_dispatch:
+
+  push:
+    branches:
+      - 'master'
+jobs:
+  tests:
+    strategy:
+      matrix:
+        platform: [ubuntu-latest]
```
idea404, line 12: Why only ubuntu here? Wouldn't you want to run this on all major OS?

test/make.test.ts
maxhr: Agreed, fixed.

.github/workflows/test-e2e.yml
maxhr: macos doesn't work yet because of this - https://github.com/near/nearcore/issues/7235

idea404 (commented):
Looks good. A couple non-blocking comments: 

1. Testing - If I read correctly, there is no test in place right now to test recommending WSL for Windows environments. Maybe good to have something like this in place. It would also be nice for the e2e tests to have a summary view once they have run, they produce a lot of output on the terminal atm which is hard to follow. 
2. Dependency versioning management - Every template comes with dependencies and each their own versions. I believe it would make the repo easier to maintain if these versions would be specified in one file and these values would be extracted when dApp is created by user. 
3. Config management / maintainability - Similar to the point above, I notice the supported languages/frameworks are specified multiple times throughout multiple files in the repo, this can get painful to manage as we add support for more languages and frameworks in the future. I would suggest using a config file and specifying these in the config file and have the rest of the repo read this file. 

Also, `npm run test` fails for the `js` tests for me in my ubuntu machine.

templates/contracts/js/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "license": "(MIT AND Apache-2.0)",
+  "type": "module",
+  "scripts": {
+    "build": "near-sdk-js build src/contract.ts build/hello_near.wasm",
+    "deploy": "npm run build && rm -rf neardev && near dev-deploy --wasmFile build/hello_near.wasm && export $(cat ./neardev/dev-account.env) && near call $CONTRACT_NAME init --accountId $CONTRACT_NAME --deposit 1 && echo $CONTRACT_NAME",
+    "test": "echo use integration-tests"
+  },
+  "dependencies": {
+    "near-cli": "^3.4.0",
+    "near-sdk-js": "0.4.0-4"
```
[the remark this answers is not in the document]
maxhr: I'd like to keep auto-generation and inter-dependency of things to minimum. I don't mind the duplication because it's a very small amount of dependencies.

src/user-input.ts
```diff
@@ -0,0 +1,189 @@
...
+  const {projectName, contract, frontend, tests} = args;
+  const hasAllOptions = contract !== undefined && frontend !== undefined;
+  const hasPartialOptions = contract !== undefined || frontend !== undefined;
+  const hasProjectName = projectName !== undefined;
+  const hasAllArgs = hasAllOptions && hasProjectName;
+  const hasNoArgs = !hasPartialOptions && !hasProjectName;
+  const optionsAreValid = hasAllOptions
+    && ['react', 'vanilla', 'none'].includes(frontend)
+    && ['js', 'rust', 'assemblyscript'].includes(contract)
+    && ['js', 'rust'].includes(tests);
```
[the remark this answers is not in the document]
maxhr: Agreed

package.json
```diff
@@ -1,68 +1,77 @@
...
+  "license": "(MIT AND Apache-2.0)",
+  "keywords": [
+    "nearprotocol",
+    "near-protocol",
+    "assemblyscript",
+    "rust",
+    "react",
+    "smart-contract",
+    "blockchain"
+  ],
```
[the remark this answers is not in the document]
maxhr: Thanks, removed some, others are false positives. The missing deps are from templates overrides folder.

gagdiez (approved):

volovyk-s (approved):

volovyk-s:
@gagdiez ok
It is better to create releases from the master. It automatically adds tags and a release here on GitHub. Now we need to do that manually.
+ do not forget to use conventional commits in this repo.

PR #1868, opened 2022-08-08
fix: remove need for specific deps-install command, use install

austinabell:
unclear to me why this isn't hooked into install. This was a footgun I ran into, devs shouldn't have to know about this command.

Tests were already broken, didn't extensively test this


maxhr (approved):
@austinabell this is great thanks!

PR #1888, opened 2022-08-18
fix: tests

maxhr:


.github/workflows/test-matrix.yml
```diff
@@ -4,7 +4,7 @@ jobs:
   tests:
     strategy:
       matrix:
-        platform: [ubuntu-latest, macos-latest, windows-latest]
```
volovyks, line 7: Do we have a plan to support all platforms?
Workspaces JS should with on macOS CI soon.

.github/workflows/test-e2e.yml
```diff
@@ -15,16 +12,17 @@ jobs:
         contract: [ js, rust, assemblyscript ]
         frontend: [ react, vanilla, none ]
         tests: [ js, rust ]
-#        exclude:
-#          - frontend: vanilla
-#          - frontend: none
+        exclude:
+        - frontend: vanilla
```
volovyks, line 16: What is the motivation to exclude some of the tests?

volovyks (approved):

.github/workflows/test-matrix.yml
maxhr: This test suite doesn't do anything platform specific, the e2e do.

PR #1889, opened 2022-08-19
fix: error message when user didnt run deploy when starting frontend

maxhr:


templates/frontend/react/frontend/near-config.js
```diff
@@ -1,5 +1,11 @@
 const CONTRACT_NAME = process.env.CONTRACT_NAME;
 
+if (!CONTRACT_NAME) {
+  const errorMessage = 'ERROR: CONTRACT_NAME not defined. Did you run deploy script (`npm run deploy`)? README.md contains more info about it.';
```
gagdiez, line 4: Here I would simply say "Did you deploy your contract?", to avoid mixing instructions for the website with instructions for the backend.

templates/frontend/react/frontend/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "license": "(MIT AND Apache-2.0)",
   "scripts": {
-    "start": "echo The app is starting! && env-cmd -f ../contract/neardev/dev-account.env parcel index.html --open",
+    "start": "echo The app is starting! && env-cmd --silent -f ../contract/neardev/dev-account.env parcel index.html --open",
```
gagdiez, line 6: Here is where I would check if the file exists and tell them if it doesn't since the user is still in the console.

maxhr: It'll make the command very complicated, and it's anyway a good idea to keep the check in client-code for the user to see it there, so that it'll be clear that he needs must have a correct contract name.

gagdiez (approved):

PR #1899, opened 2022-08-25
fix: moved package.json commands to bash scripts

maxhr:
- add start.sh for frontends to check dev account file
- add build.sh + deploy.sh to contracts so that it'll be easier to understand what it's doing
- generated package.json now runs build+deploy scripts
- moved files in frontend template dir to simplify folder structure


templates/frontend/shared/start.sh
```diff
@@ -0,0 +1,23 @@
...
+  echo The app is starting!
+  env-cmd -f $DEV_ACCOUNT_FILE parcel index.html --open
+}
+
+alert () {
+  echo "======================================================"
+  echo "Can't find the dev-account credentials file (${DEV_ACCOUNT_FILE})"
+  echo "Did you deploy your contract ?"
+  echo ">> Run 'npm run deploy' from your 'contract' directory"
+  echo "======================================================"
```
maxhr, line 16: @gagdiez wdyt about the message

templates/contracts/js/deploy.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+./build.sh
+
+echo ">> Deploying contract"
+
+# Deploy the contract with an auto-generated development account
+rm -rf neardev
```
gagdiez, line 8: we can delete this line since we do not have it in all deploy.sh

templates/contracts/js/deploy.sh
```diff
@@ -0,0 +1,16 @@
...
+# https://docs.near.org/tools/near-cli#near-dev-deploy
+near dev-deploy --wasmFile build/hello_near.wasm
+
+export $(cat ./neardev/dev-account.env)
+
+near call $CONTRACT_NAME init --accountId $CONTRACT_NAME --deposit 1
+
+echo $CONTRACT_NAME
```
gagdiez, line 16: same as before

templates/contracts/js/deploy.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+./build.sh
+
+echo ">> Deploying contract"
+
+# https://docs.near.org/tools/near-cli#near-dev-deploy
+near dev-deploy --wasmFile build/hello_near.wasm
+
+export $(cat ./neardev/dev-account.env)
```
gagdiez, line 10: Now that I think about it, we could better use `export CONTRACT_NAME=$(cat ./neardev/dev-account)` since it is more explicit in this case

templates/frontend/shared/start.sh
gagdiez: I would recommend the 'root' directory, since people will be most probably there, and the `package.json` is ready for that.

gagdiez (approved):

templates/frontend/shared/start.sh
gagdiez: suggestion: 

```bash
GREEN='\033[1;32m'
NC='\033[0m' # No Color

echo "======================================================"
echo "It looks like you forgot to deploy your contract"
echo ">> Run ${GREEN}'npm run deploy'${NC} from the 'root' directory"
echo "======================================================"
```

PR #1901, opened 2022-08-27
feat: better compatibility

Olyno:
# Description

This pull request adds more compatibility with existing tools. Templates become monorepos, and now support the other popular package managers, namely ``yarn'`` and ``pnpm``.

The ``.sh`` files are not compatible with Windows, so I changed to a "scripts" folder instead, which gathers the different "sh" scripts but in Javascript (making them compatible with Windows).

**tltr:**
 - Move ``.sh`` scripts to a ``scripts`` folder, and change the format to JS
 - Add [yarn 2+](https://yarnpkg.com/) and [pnpm](https://pnpm.io/) support
 - Switch template to monorepos, removing ``package-lock.json`` files

***Note:** I also renamed ``contract`` to ``contracts`` instead, which make more sense to me. This is something optional that i can revert if you won't**


maxhr:
Thanks @Olyno for the contribution! We'll get to reviewing it this week.
First comment - we don't want to introduce an additional prompt question to the users. Advanced users should be able to choose their package managers but newbies shouldn't be aware that there is such thing.
How do you think we should solve it?

gagdiez:
This branch has conflicts, and there were no other community members asking for this compatibility option, for this, I will now close the PR.

PR #1913, opened 2022-09-20
Release 5.1.0

gagdiez:
Updated create-near-app with the latest "Hello World" from the near-examples


maxhr (approved):
@gagdiez can you update snapshots again after the minor fix?
I re-ran the 2e2 manually tests because sometimes they fail randomly when trying to download deps.

PR #1940, opened 2022-11-10
link in /rust/src/lib.rs was deprecated. updated to latest link

thalasith:
Updated link as the last link led to a 404 error. Hopefully this well help future developers!


frol:
@thalasith Thanks for submitting this PR!

The docs were moved to https://docs.near.org, and the file is already updated in master, so I close this PR.

PR #2014, opened 2023-06-21
Build contract before integration tests

agostbiro:
The integration tests depend on a built artifact, so they'll fail unless the `build:contract` command is ran manually before running integration tests.

It's also error-prone as users might not realize that they have to rebuild manually before running integration tests, and they might get confused.

Incidentally this change also fixes the e2e tests that were failing in master locally on a fresh checkout of the repository.

Branched out from https://github.com/near/create-near-app/pull/2013, otherwise tests are failing.

### DIFF
--- a/src/package-json.ts
+++ b/src/package-json.ts
@@ -44,15 +44,17 @@ const buildScript = (hasFrontend: boolean): Entries => hasFrontend ? {
   'build': 'npm run build:contract',
 };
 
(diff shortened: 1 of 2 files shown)


frol (approved):

frol:
Re-opening this PR to trigger CI

gagdiez (approved):

PR #2017, opened 2023-06-27
Remove ledger to fix ci

gagdiez:


frol (approved):
It is sad, but it is better to have somewhat working tool than completely broken, so let's merge it

PR #2018, opened 2023-06-27
Update dependencies

gagdiez:


test/__snapshots__/make.test.ts.snap
```diff
@@ -254,84 +254,84 @@ and then use the logged account to sign the transaction: \`--accountId <your-acc
...
-  \\"name\\": \\"hello_near\\",
-  \\"version\\": \\"1.0.0\\",
-  \\"license\\": \\"(MIT AND Apache-2.0)\\",
-  \\"type\\": \\"module\\",
-  \\"scripts\\": {
-    \\"build\\": \\"./build.sh\\",
-    \\"deploy\\": \\"./deploy.sh\\",
-    \\"test\\": \\"echo use integration-tests\\"
+  "name": "hello_near",
+  "version": "1.0.0",
```
frol, line 305: I wonder if this actually works (quotes not being escaped feels wrong to me)

gagdiez: those files are generated automatically, I do not think we should manually tinker with them

frol (approved):
I hope you know what you are doing 😄 

The change looks fine to me

test/__snapshots__/make.test.ts.snap
frol: I agree, but it looks odd to me, so I just flagged it in case that would trigger some alarm in your head

PR #2114, opened 2023-11-17
refactor: change the npm test script to build the ts contract

flmel:
The `test` script in package.json currently requires the user to manually build the contract in prior to running the tests.

Ideally the flow would be the same as the `test.sh` script for the rust smart contract where the test script builds the contract prior execution.

### DIFF
--- a/templates/contracts/ts/package.json
+++ b/templates/contracts/ts/package.json
@@ -9,7 +9,7 @@
   "scripts": {
     "build": "near-sdk-js build src/contract.ts build/hello_near.wasm",
     "deploy": "near dev-deploy --wasmFile build/hello_near.wasm",
-    "test": "npm run build && cd sandbox-ts && $npm_execpath run test -- -- ../build/hello_near.wasm",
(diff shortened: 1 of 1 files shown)


templates/contracts/ts/package.json
```diff
@@ -9,7 +9,7 @@
   "scripts": {
     "build": "near-sdk-js build src/contract.ts build/hello_near.wasm",
     "deploy": "near dev-deploy --wasmFile build/hello_near.wasm",
-    "test": "cd sandbox-ts && $npm_execpath run test -- -- ../build/hello_near.wasm",
+    "test": "npm run build && cd sandbox-ts && $npm_execpath run test -- -- ../build/hello_near.wasm",
```
gagdiez, line 12: lets use `$npm_execpath` so the user can choose their own package manager (e.g. npm, yarn, pnpm, etc)

gagdiez (changes requested):

gagdiez (approved):

PR #2117, opened 2023-12-11
Nextjs template structure enhancement

friendlymatthew:
This project enhances the existing `create-near-app` Next.js template by restructuring its directory and file organization. The goal was to align more closely with standard Nextjs conventions.

## The Problem
The [original](https://github.com/near/create-near-app/tree/master/templates/frontend/next) template adopts a project structure that deviates from common Next.js practices.

For example:
- Usage of `page.js` as a filename is not standard in Next.js projects
- The global HTML structure was directly in the `RootLayout` component
- Global CSS resided in the `RootLayout` component
- Why is `favicon.ico` in the `app/` directory?

See the original structure here:
```
| next/
    | public/
    | src/
        | app/
            | hello-components/
                | page.js
            | hello-near/
                | page.js
            | app.module.css
            | favicon.ico
            | globals.css
            | layout.js
            | page.js
        | components/
        | wallets/
        | config.js
```

## The Solution
I restructured the project to more closely follow Nextjs conventions. Afterward, I noticed [`near-discovery`](https://github.com/near/near-discovery/tree/develop/src) follow similar conventions.

The new structure is: 
```
| create-near-app/
    | public/
        | favicon.ico
    | src/
        | layout.js
        | pages/
            | index.js
            | hello-components/
                | index.js
            | hello-near/
                | index.js
            | _app.js
            | _document.js
        | components/
        | styles/
            | app.module.css
            | globals.css
        | wallets/
        | config.js
```

### Changes Proposed
- file renaming: renamed `page.js` to `index.js` for clarity
- directory restructuring: moved all page-related files into a dedicated `pages/` directory
- styles organization: centralized `.css` files in the `styles/` directory
- app level file: introduced `_app.js` for app-level configurations and wrappers
- layout extraction: moved global layout logic from `layout.js` to `_document.js`

## Why I did this
- It is more intuitive
- I spent my Sunday morning picking my hair because I wanted to set up `next-auth` but the original template was too subjective


gagdiez:
hi @friendlymatthew, please notice that `create-near-app` is following the new default convention for `next.js` apps, you can learn more about it here: https://nextjs.org/docs#app-router-vs-pages-router. This is becase `create-near-app` comes from a changing as little as possible a base `create-next-app`.

Now, a valid discussion is to think if we want to stick to the old pages router, since more people might know it, or with the new app router, since anyway that is the direction of `next.js`.

I personally prefer to stick with the new router, since that way we will have less technical debt in the future.

gagdiez (changes requested):
See main comments in the `Conversation` tab.

flmel:
I support sticking to the new app router. Less tech dept, forward looking, also people shall be up to date of the new changes to next tooling or if they are not, create-near-app can be a heads up for that

friendlymatthew:
As I look up at your comments @flmel @gagdiez, I agree. Thanks for considering. FWIW, it would be cool to have the option to choose. Nextjs and `create-next-app` provide the option to choose so that chunk of coals like me that prefer the old method can continue hacking. But since that would mean more maintaining of code, I understand.


Screen Shot 2023-12-12 at 1 48 33 PM

gagdiez:
@friendlymatthew I would expect the app to change fairly little in the future, we could let the user select their preferred router.

Do you feel this will fix more problems than the ones it will bring? (imagine you have to maintain it) if yes, I'm ok with implementing the router selector

gagdiez:
this would mean creating two frontend templates: `next-app` and `next-pages`, and adding a selection during the user's input

friendlymatthew:
> Do you feel this will fix more problems than the ones it will bring? (imagine you have to maintain it) if yes, I'm ok with implementing the router selector

Here are some of my thoughts: 

### **The current state of Next.js**
We're in this transitory phase where people are *slowly* migrating to the app router. Yes, the app router is forward looking but it hasn't been battle tested like the pages router. 

Most importantly, the surrounding ecosystem hasn't caught up. A lot of dependencies won't support the app router yet. The reason why I put this PR up was because I was struggling to integrate `next-auth`. 

Plus, a lot of documentation and videos still reference the pages directory, so I imagine newer devs may struggle more.

<br />

### **Weighing the downsides**
Building another template will introduce maintenance overhead, but I don't foresee a lot, if any. According to Nextjs docs: *the pages router is still supported in newer versions of Next.js*. So even though it's recommended to use the app router, we won't be facing any backward compatibility issues.

If the router selector gets built out and some problems occur, I'll remember this hill that I'm fighting on and gladly help

friendlymatthew:
> this would mean creating two frontend templates: `next-app` and `next-pages`, and adding a selection during the user's input

Yup -- if we agree to proceed, gladly help out and tack a commit onto this PR

gagdiez:
@friendlymatthew let's go ahead and do this. I think it is valuable for the ecosystem. I'll be back on vacations from January 1st and can help with the implementation as well after that date.

friendlymatthew:
> @friendlymatthew let's go ahead and do this. I think it is valuable for the ecosystem. I'll be back on vacations from January 1st and can help with the implementation as well after that date.

Sounds good -- I'll make some time around Jan 2nd for this feature. FWIW, I've looked at the internals and it shouldn't be hard to implement.

Have a good vacation!

gagdiez:
hey @friendlymatthew, I am closing this since we **merged your contributions** in #2127 . Will test it now throughly and make a release, thanks!

friendlymatthew:
> hey @friendlymatthew, I am closing this since we **merged your contributions** in #2127 . Will test it now throughly and make a release, thanks!

Awesome

PR #2157, opened 2024-09-02
Update the starter template

gtsonevv:


gagdiez:
Hi @gtsonevv 

thanks for the contribution, but I do not feel that the FT and NFT examples (specially written on ts, specially without tests) are a good starting point for developers

what was the main idea behind this contribution? I assume to add more "starting" points for everyone. If that's the case, we might want to change the tool into something different, where people choose what type of smart contract to use as starting point.

This would require to add all contracts in both Rust and JS, with proper testing, as well as to change the menu so people can select

We would also need to collaborate with @frol to propagate this to `cargo near new`

Let me know if you need any guidance on doing these changes!

gtsonevv:
Hey @gagdiez, yes, the idea was to provide developers with ready-to-use implementations of key NEAR contract standards by adding examples of those contracts.

gagdiez:
in `v9` we have decided to start using `auction` and `auction advance` as starter templates, we are evaluating if providing also `FT` and `NFT`, but will do so through the `near-sdk-contract-tools`

PR #2175, opened 2025-05-23
Download rust contract template from cargo-near when it's possible

garikbesson:


garikbesson:
After discussing it with @gagdiez and reviewing usage metrics, we decided to disable creating smart contracts in Rust as a feature and show warning to guide users to cargo-near github page instead of removing choosing options to create smart contracts and creating TypeScript smart contract by default. To do that I opened another [PR](https://github.com/near/create-near-app/pull/2176)

